PR #10, opened 2018-11-12
removed System.out.println()

symphony-jeremy:
Hi @manucaicedo 

this is a minor PR to remove debug outputs from System.out. I think we can also consider using the logger to display exception stack traces WDYT ?

I started using this lib to prepare the hackathon, it's super useful to be quickly set up ! 👍


symphony-jeremy:
I just noticed that this small old PR has been forgotten. Can you merge it or close it @mistryvinay @opoupeney ?

PR #20, opened 2019-01-18
Allow PresentationML to be sent

anthony-symphony:
Fixed private sendMessage for when appendTags is false
Above fixes forwardMessage so it works
Check and appendTags only if necessary in public sendMessage function


anthony-symphony:
PresentationML detection not perfect.

PR #22, opened 2019-02-06
Changed SymBotRSAAuth to use proxy for KM Auth

anthony-symphony:
KM Auth for RSA was not using proxyClient. Changed it to match cert auth code.


anthony-symphony:
Added variables for keyManagerProxy as well as modified Certificate Auth to respect those variables as well.

PR #24, opened 2019-02-14
fix of StreamInfo

ronerjr:
This pull request contains a fix on StreamInfo.
The API is sending an object on payload, not a string. In order to make the things easier I created an enumerator to be used on comparisons.

### DIFF
--- a/src/main/java/model/StreamInfo.java
+++ b/src/main/java/model/StreamInfo.java
@@ -9,7 +9,7 @@ public class StreamInfo {
     private String origin;
     private Boolean active;
     private Long lastMessageDate;
-    private String streamType;
+    private StreamType streamType;
     private StreamAttributes streamAttributes;
     private RoomName roomAttributes;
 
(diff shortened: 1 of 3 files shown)


ystan-:
Can StreamType be used directly as an enum in the method signatures rather than as a wrapper class?

ronerjr:
> Can StreamType be used directly as an enum in the method signatures rather than as a wrapper class?

Can you give me an example?
It's is a model to parse the return of api https://rest-api.symphony.com/v1.53.1/reference#stream-info-v2 that returns an object of StreamType, them I put an enum to help us on future comparisons avoiding mistakes with String.

ystan-:
You created StreamType as a wrapper class to contain a single field of the new enum type StreamTypes, so I'm wondering what the considerations were behind this. As opposed to using the new enum directly in StreamInfo. This will be a breaking change, so it should be as clean as possible.

For example, remove StreamType, rename StreamTypes as StreamType, leave StreamInfo to reference the new StreamType.

`package model;

public enum StreamType~~s~~ {
  IM, ROOMS
}`

ronerjr:
@ystan- Here an example of the API response:
` "streamType": {
    "type": "IM"
  },`
Core team exposed the StreamType as an object, so we need to create one of our side to parse the json tree to java object, this is why I didn't use the enum directly in StreamInfo.

ystan-:
my bad. this can go.

PR #35, opened 2019-03-03
Datafeed read handle errors

useralan72:
error should not throw null but empty list and log error


ystan-:
Is this a duplicate of PR #34 ?

useralan72:
Closing off as duplicate of #34

PR #51, opened 2019-05-07
fix proxy support for ext apps

marcus-secato-symphony:
Fix for proxy support for extension apps. Previously it was not taking into account the podProxyUrl property for extension apps.


mistryvinay:
We should also update the SymBotAuth client as well, so it matches these latest udpates. https://github.com/marcus-secato-symphony/symphony-api-client-java/blob/master/src/main/java/authentication/SymBotAuth.java

Issue #58, opened 2019-06-07
SymExtensionAppRSAAuth only supports app RSA key specified in SymConfig


davidelorenzoli:
SymExtensionAppRSAAuth only allows to specify the app RSA key in SymConfig as a path to the RSA file. This is an issue in case the RSA key is hosted somewhere not reachable by the supported SymConfig path. Or if the RSA key is generated on the fly.

A more flexible way, to complement SymConfig, is to pass to SymExtensionAppRSAAuth a PrivateKey object so that SymExtensionAppRSAAuth does not need to worry to support all possible ways a private key is fetched.

PR #59, opened 2019-06-07
(#58) Added constructor that takes as argument the RSA PrivateKey object

davidelorenzoli:
SymExtensionAppRSAAuth only allows to specify the app RSA key in SymConfig as a path to the RSA file. This is an issue in case the RSA key is hosted somewhere not reachable by the supported SymConfig path. Or if the RSA key is generated on the fly.

A more flexible way, to complement SymConfig, is to pass to SymExtensionAppRSAAuth a PrivateKey object so that SymExtensionAppRSAAuth does not need to worry to support all possible ways a private key is fetched.

### DIFF
--- a/src/main/java/authentication/SymExtensionAppRSAAuth.java
+++ b/src/main/java/authentication/SymExtensionAppRSAAuth.java
@@ -1,8 +1,5 @@
 package authentication;
 
-import static org.apache.commons.lang3.StringUtils.isEmpty;
-import static utils.JwtHelper.validateJwt;
-
 import authentication.extensionapp.InMemoryTokensRepository;
 import authentication.extensionapp.TokensRepository;
 import clients.symphony.api.APIClient;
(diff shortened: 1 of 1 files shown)


src/main/java/authentication/SymExtensionAppRSAAuth.java
```diff
@@ -19,6 +16,11 @@
 import utils.HttpClientBuilderHelper;
 import utils.JwtHelper;
 
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
```
artbegolli: Why are these reshuffled?

src/main/java/authentication/SymExtensionAppRSAAuth.java
```diff
@@ -122,15 +142,17 @@ public AppAuthResponse appAuthenticate() {
...
     }
 
-    private PrivateKey getPrivateKey() {
-        PrivateKey privateKey = null;
-        try {
-            privateKey = JwtHelper.parseRSAPrivateKey(
-                new File(config.getAppPrivateKeyPath() + config.getAppPrivateKeyName()));
-        } catch (IOException | GeneralSecurityException e) {
-            logger.error(e.getMessage());
+    private PrivateKey getAppPrivateKey() {
```
bgguna: Even though this is not  in scope of the PR, this is a private method so it should not have any impact.
Moreover, it gives actually context to the private key.

PR #74, opened 2019-09-04
Enhancement to allow Bot to listen to own events on a per listener basis.

anthony-symphony:


symphony-youri:
@ystan- Do you remember of issues related to this change? (or maybe you closed it because of the lack of activity)

ystan-:
@symphony-youri this change didn't fit in with how we expected the sdk to be used back then. feel free to revisit with bdk 2.0.

PR #86, opened 2020-02-05
PLAT-8645 Added missing params (skip and limit) to streams list

symphony-thibault:
Related endpoint : [#list-user-streams](https://developers.symphony.com/restapi/reference#list-user-streams)

### Purpose

When using the `StreamsClient.getUserStreams` function to list the user streams, there is no provision to allow the user to pass the `skip` or `limit` parameter, making the function limited to returning the 50 first rooms.

### What has been done

Prototype of the existing `getUserStreams(List<String>, boolean)` has been preserved for backward compatibility purpose. An additional method has been introduced including the `limit` and `skip` parameters : 

```java
public List<StreamListItem> getUserStreams(List<String>, boolean, int skip, int limit);
```

This method also checks that those parameters are not lower than 0.

### DIFF
--- a/src/main/java/clients/symphony/api/StreamsClient.java
+++ b/src/main/java/clients/symphony/api/StreamsClient.java
@@ -306,6 +306,7 @@ public class StreamsClient extends APIClient {
      * @param includeInactiveStreams Whether to include inactive conversations.
      * @return a list of all the streams
      * @throws SymClientException the generic client exception
+     * @throws IllegalArgumentException on illegal skip or limit parameter
      */
     public List<StreamListItem> getUserStreams(List<String> streamTypes, boolean includeInactiveStreams)
         throws SymClientException {
@@ -322,6 +323,7 @@ public class StreamsClient extends APIClient {
      * @param limit Maximum number of streams to return. If 0, all user streams will be returned.
      * @return a list of all the streams
      * @throws SymClientException the generic client exception
+     * @throws IllegalArgumentException on illegal skip or limit parameter
      */
     public List<StreamListItem> getUserStreams(List<String> streamTypes, boolean includeInactiveStreams, int skip, int limit)
         throws SymClientException {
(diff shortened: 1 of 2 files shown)


src/main/java/clients/symphony/api/StreamsClient.java
```diff
@@ -294,8 +294,48 @@ public RoomSearchResult searchRooms(RoomSearchQuery query, int skip, int limit)
...
+     *
+     * @param streamTypes A list of stream types that will be returned.
+     * @param includeInactiveStreams Whether to include inactive conversations.
+     * @param skip Number of stream results to skip.
+     * @param limit Maximum number of streams to return. If 0, all user streams will be returned.
+     * @return a list of all the streams
+     * @throws SymClientException the generic client exception
+     * @throws IllegalArgumentException on illegal skip or limit parameter
+     */
+    public List<StreamListItem> getUserStreams(List<String> streamTypes, boolean includeInactiveStreams, int skip, int limit)
```
symphonydarlys: do we have max limit?

src/test/java/it/clients/symphony/api/StreamsClientTest.java
```diff
@@ -485,4 +487,14 @@ public void getUserStreamsSuccess() {
...
+
+  @Test(expected = IllegalArgumentException.class)
+  public void failToGetUserStreamsWithIllegalSkipValue() {
+    this.streamsClient.getUserStreams(Collections.emptyList(), true, -1, 0);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void failToGetUserStreamsWithIllegalLimitValue() {
+    this.streamsClient.getUserStreams(Collections.emptyList(), true, 0, -1);
+  }
```
symphonydarlys: Is it possible to add a new test to cover the conventional behavior of the new method?

symphony-thibault: It is actually already covered

PR #96, opened 2020-03-02
Fallback to API call if sessionToken does not contain userid or username

anthony-symphony:


src/main/java/clients/SymBotClient.java
```diff
@@ -168,10 +168,17 @@ private SymBotClient(SymConfig config, ISymAuth symBotAuth) {
...
 
-        this.botUserInfo = parseUserFromSessionToken(symBotAuth.getSessionToken());
-
+        try {
+            this.botUserInfo = parseUserFromSessionToken(symBotAuth.getSessionToken());
+        }
+        // Exception happens when you can not get userId or username out of the JWT
+        // When that happens grab account info via API.
+        catch (ArrayIndexOutOfBoundsException e) {
+            getBotUserInfo();
```
symphony-thibault: Add a debug log before this line, saying that something like "Using API to retrieve user info"

src/main/java/clients/SymBotClient.java
```diff
@@ -218,8 +225,26 @@ private void reportIfLoadBalanced(SymConfig config) {
...
+        }
+        return botUserInfo.getUsername();
+    }
+    
+    public long getBotUserId() {
+        if (botUserInfo == null || botUserInfo.getId() == null) {
+            getBotUserInfo();
+        }
+        return botUserInfo.getId();
+    }
```
symphony-thibault: `SymBotClient` implements the `ISymClient` interface, are you sure that you don't want to move those new methods at interface level?

anthony-symphony: These changes are specific to bots so should not be in ISymClient

symphony-thibault:
@anthony-symphony Don't forget to retrofit this change to https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/tree/1.0-rc

PR #99, opened 2020-03-11
SDK-38 - Fixing issues with addTextField and addTextArea methods

lmacedodaitan:


src/main/java/utils/FormBuilder.java
```diff
@@ -97,13 +100,33 @@ public FormBuilder addTextField(String name, String display, String placeholder,
...
+            .addField("placeholder", placeholder)
+            .addField("required", required);
+
+        if (display != null) {
+            messageML.append(textAreaML.setContents(display).build());
+        } else {
+            messageML.append(textAreaML.buildSelfClosing());
+        }
 
         messageML.append(textAreaML);
```
symphony-enrico: I should not be removed?

src/main/java/utils/FormBuilder.java
```diff
@@ -97,13 +100,33 @@ public FormBuilder addTextField(String name, String display, String placeholder,
         return this;
     }
 
-    public FormBuilder addTextArea(String name, String display, String placeholder, boolean required) {
-        String textAreaML = TagBuilder.builder("textarea")
+    public FormBuilder addTextField(String name, String display, String placeholder, boolean required, boolean masked) {
+        TagBuilder textFieldML = TagBuilder.builder("text-field")
```
symphony-thibault: Are you sure that modifying this method prototype doesn't break backward compatibility with previous versions? Reminder: we have to be very careful with SDKs public methods that might be already in use by some of our customers. 

If there is a possibility for this method to be use, `@Deprecated` it and create another one with the additional parameter `boolean masked`

lmacedodaitan: @symphony-thibault are we good with these changes?

PR #101, opened 2020-03-18
SDK-17 Fixed external LoadBalanced config 

symphony-thibault:
The external LoadBalanced was broken or has never worked as expected. This PR is more a patch that a real clean fix. 

#### `SymConfig.java`

The method `getAgentUrl()` was using field `this.agentHost` to compute the full Agent server url. Instead, it has to use `this.getAgentHost()` because this method is overridden in `SymLoadBalancedConfig.java`. 

#### `SymLoadBalancedConfig.java`

The way that the method `public String getActualAgentHost()` has been implemented was wrong : 
```java
String externalAgentHost = (agentServers != null && agentServers.size() > 0) ?
            agentServers.get(0) : getAgentHost();
```
It was apparently trying to get LB host from the Agent servers defined in the `config-lb.json` file. Moreover, having `agentServers` empty could have lead to a never-ending recursive loop with de subsequent call to `getAgentHost()`...


symphonydarlys:
LGTM

PR #109, opened 2020-05-14
Adding missing RoomSearchQuery parameters: sortOrder and subType

lmacedodaitan:


symphonydarlys:
This is typical example where we have to add integration test to validate this change.

symphonydarlys:
How validate this change? @lmacedodaitan has you created a bot to test result with these new parameters?

lmacedodaitan:
> How validate this change? @lmacedodaitan has you created a bot to test result with these new parameters?

Yes Darlys, before I opened the PR, I tested using a bot pointing to a pod.

src/main/java/model/RoomSearchQuery.java
```diff
@@ -12,6 +12,8 @@
     private NumericId creator;
     private NumericId owner;
     private NumericId member;
+    private String sortOrder;
```
symphony-thibault, line 15: Could you also add `@JsonIgnoreProperties(ignoreUnknown = true)` on top of the class? So that if another field is added in the future on the server side it won't break the SDK.

PR #110, opened 2020-05-18
PLAT-9037 Adding 'quickws' to be accepted as a valid uri scheme

lmacedodaitan:
Bug details on https://perzoinc.atlassian.net/browse/PLAT-9037
I'm adding the 'quickws' to be accepted as a valid URI scheme on Java SDK implementation of IDataProvider's validateURI method.

### DIFF
--- a/src/main/java/utils/DataProvider.java
+++ b/src/main/java/utils/DataProvider.java
@@ -22,6 +22,7 @@ public class DataProvider implements IDataProvider {
     public DataProvider(SymBotClient botClient) {
         STANDARD_URI_SCHEMES.add("http");
         STANDARD_URI_SCHEMES.add("https");
+        STANDARD_URI_SCHEMES.add("quickws");
         this.botClient = botClient;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/utils/DataProvider.java
```diff
@@ -22,6 +22,7 @@
     public DataProvider(SymBotClient botClient) {
         STANDARD_URI_SCHEMES.add("http");
         STANDARD_URI_SCHEMES.add("https");
+        STANDARD_URI_SCHEMES.add("quickws");
```
symphony-thibault, line 25: Hum, that sounds very specific to be added like that to the SDK. It would be better to have a new configuration property called `"supportedUriSchemes"` which would have as default value `["http", "https"]`. Then the bot developer could decide to add new ones like : 
```json
{
    ...
    "supportedUriSchemes": ["http", "https", "quickws"]
}
```
Screenshot 2020-05-18 at 18 34 23

PR #113, opened 2020-05-22
CES-898: Created an API to get message by id along with the disclaimer text

alokpatel11290:
## Description
- [Jira Ticket](https://perzoinc.atlassian.net/browse/CES-898)

## Implementation
- We have created an API to get the message using the message id
- Added Unit Test for the same API
- Added `sharedMessage`, `disclaimer` & `sid` fields to the `InboundMessage` following V4Message structure on the agent


alokpatel11290:
@symphony-thibault You are welcome!
I will create a PR against the release branch also.
Thank you for the quick response. 👍

PR #115, opened 2020-05-28
PLAT-9069: Handle null for bad data structure to getMentions

symphony-hong:
Add more filters to handle bad structure of received data.


symphony-thibault:
@symphony-hong could you also backport this change to branch `1.1-rc`?

PR #119, opened 2020-06-02
Map and expose formMessageId

anthony-symphony:


src/main/java/model/events/SymphonyElementsAction.java
```diff
@@ -53,4 +54,12 @@ public void setFormId(String formId) {
     public void setFormValues(Map<String, Object> formValues) {
         this.formValues = formValues;
     }
+    
+    public String getFormMessageId() {
+        return formMessageId;
+    }
+
+    public void setFormMessageId(String formValues) {
```
symphony-hong, line 62: Should change `String formValues` to `String formMessageId`

PR #126, opened 2020-06-10
SDK-59 Add Initiator to RoomListener callback

symphony-hong:


src/main/java/listeners/RoomListener.java
```diff
@@ -1,17 +1,18 @@
...
-    void onUserLeftRoom(UserLeftRoom userLeftRoom);
+    void onRoomMessage(InboundMessage message, Initiator initiator);
+    void onRoomCreated(RoomCreated roomCreated, Initiator initiator);
+    void onRoomDeactivated(RoomDeactivated roomDeactivated, Initiator initiator);
+    void onRoomMemberDemotedFromOwner(RoomMemberDemotedFromOwner roomMemberDemotedFromOwner, Initiator initiator);
+    void onRoomMemberPromotedToOwner(RoomMemberPromotedToOwner roomMemberPromotedToOwner, Initiator initiator);
+    void onRoomReactivated(Stream stream, Initiator initiator);
+    void onRoomUpdated(RoomUpdated roomUpdated, Initiator initiator);
+    void onUserJoinedRoom(UserJoinedRoom userJoinedRoom, Initiator initiator);
+    void onUserLeftRoom(UserLeftRoom userLeftRoom, Initiator initiator);
```
symphony-thibault, line 17: Here you introduced a breaking in the API. We have to be very careful with that in our following developments! 

2 solutions : 
- mark existing method as deprecated and duplicate it with second `Initiator` param
- use varargs for `Initiator` param and check in the method if it's present or not : 

```java
void onRoomMessage(InboundMessage message, Initiator... initiator);
```

symphony-thibault: (I actually prefer the first solution, `@Deprecated` + second method with `Initiator` param)

symphony-hong: Since the customer only requested for userjoinedroom event, I have an idea that we can create an Initiator field in UserJoinedRoom class and set it before giving to the listener.
`UserJoinedRoom userJoinedRoom = event.getPayload.getUserJoinedRoom;`
`userJoinedRoom.setInitiator(initiator);`
`listener.onUserJoinedRoom(userJoinedRoom);`
Is it a solution?

symphony-thibault: Don't you think that we could actually use the same strategy for other callbacks?

src/main/java/listeners/RoomListener.java
```diff
@@ -1,6 +1,7 @@
 package listeners;
 
 import model.InboundMessage;
+import model.Initiator;
```
symphony-thibault, line 4: Unused import?

symphony-hong: yeah I will remove it

src/main/java/model/events/MessageSent.java
```diff
@@ -2,14 +2,23 @@
...
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 
 public class MessageSent {
 
     private InboundMessage message;
 
+    private Initiator initiator;
 
+    public Initiator getInitiator() {
```
symphony-elias, line 15: Since same code is duplicated across several classes, I would suggest having a base abstract class containing the Initiator field and the corresponding getter and setter.

PR #127, opened 2020-06-10
Added unittest for SDK-63

symphony-elias:
check unknown fields returned by healtcheck api are ignored

### DIFF
--- a/src/test/java/it/clients/symphony/api/HealthcheckClientTest.java
+++ b/src/test/java/it/clients/symphony/api/HealthcheckClientTest.java
@@ -3,11 +3,14 @@ package it.clients.symphony.api;
 import clients.symphony.api.HealthcheckClient;
 import clients.symphony.api.constants.AgentConstants;
 import it.commons.BotTest;
+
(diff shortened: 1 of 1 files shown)


src/test/java/it/clients/symphony/api/HealthcheckClientTest.java
```diff
@@ -43,4 +46,27 @@ public void performHealthCheckSuccess() {
     assertEquals(true, response.getPodConnectivity());
   }
 
+  @Test
+  public void performHealthCheckSuccessWithUnknownField() {
+    stubFor(get(urlEqualTo(AgentConstants.HEALTHCHECK))
```
symphony-thibault, line 51: Could you consider using utility method for mocking HTTP calls : https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/src/test/java/it/commons/BotTest.java#L48 ?

symphony-elias: Thanks, I made the changes.

PR #132, opened 2020-06-17
SDK-90: Handle null content-type 401 error response

symphony-hong:
The issue in ticket SDK-90 comes from the 401 error returned with no content-type in headers.
`ClientError error = response.readEntity(ClientError.class);`
So, instead using Json parser to `readEntity` as `ClientError` object, we will content of response as a String then parse it with Jackson.
```
String errorString = response.readEntity(String.class);
ObjectMapper mapper = new ObjectMapper();
ClientError error;
try {
       error = mapper.readValue(errorString, ClientError.class);
} catch (JsonProcessingException e) {
       throw new APIClientErrorException(e.getMessage());
}
```

### DIFF
--- a/src/main/java/clients/symphony/api/APIClient.java
+++ b/src/main/java/clients/symphony/api/APIClient.java
@@ -14,6 +14,17 @@ public abstract class APIClient {
     private final Logger logger = LoggerFactory.getLogger(APIClient.class);
     private static final ObjectMapper mapper = new ObjectMapper();
 
+    private ClientError readClientError(Response response) {
+        String errorString = response.readEntity(String.class);
+        try {
+            return mapper.readValue(errorString, ClientError.class);
+        } catch (JsonProcessingException e) {
+            ClientError error = new ClientError();
+            error.setMessage(errorString);
+            return error;
+        }
+    }
+
     protected void handleError(Response response, ISymClient botClient) throws SymClientException {
         if (response.getStatusInfo().getFamily() == Family.SERVER_ERROR) {
             logger.error("REST error: error code {} reason {}",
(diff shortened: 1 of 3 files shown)


src/main/java/clients/symphony/api/APIClient.java
```diff
@@ -19,7 +21,15 @@ protected void handleError(Response response, ISymClient botClient) throws SymCl
             );
             throw new ServerErrorException(response.getStatusInfo().getReasonPhrase());
         } else {
-            ClientError error = response.readEntity(ClientError.class);
+            String errorString = response.readEntity(String.class);
```
symphony-elias, line 24: Could we try to set a Content-Type header to application/json if it does not exist ?

Or at least extract the parsing into a dedicated function. And put the ObjectMapper as static field since it is quite a heavy object designed for reuse, see documentation : https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#ObjectMapper()

symphony-hong: I tried to find way to set Content-Type header for response but I didn't find anything. Not sure if it's possible or not.
Maybe I will try the second suggestion.

src/main/java/clients/symphony/api/APIClient.java
```diff
@@ -19,7 +22,14 @@ protected void handleError(Response response, ISymClient botClient) throws SymCl
...
+            ClientError error;
+            try {
+                error = mapper.readValue(errorString, ClientError.class);
+            } catch (JsonProcessingException e) {
+                throw new APIClientErrorException(e.getMessage());
```
symphony-thibault, line 30: Do we really want to return the `JsonProcessingException` error message or directly `errorString` value? For some raisons it would be interesting to know what has actually been returned by the server, and this error might not be a Json object (it happens with SBE )

PR #143, opened 2020-06-22
SDK-104: Adding fields in config.json for specifying context path

symphony-hong:


src/test/java/it/configuration/SymConfigTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static org.junit.Assert.*;
+
+public class SymConfigTest {
+
+    private SymConfig config;
+
+    @Before
+    public void setUp() {
+        InputStream configStream = SymConfigTest.class.getResourceAsStream("/bot-config.json");
```
symphony-elias, line 18: Was the bot-config.json modified for the tests? If so, please put the changes.

symphony-hong: No, I did not modify the bot-config.json because if I do so it will break all the others test.

symphony-hong: I use this before each test to set the context path `config.setContextPath("ctx");`

PR #148, opened 2020-06-26
Close FileInputStream after reading in FileHelper

anthony-symphony:
Close FileInputStream after reading in FileHelper. Test fails due to trying to delete the file while still in use.

### DIFF
--- a/src/main/java/internal/FileHelper.java
+++ b/src/main/java/internal/FileHelper.java
@@ -32,8 +32,15 @@ public class FileHelper {
     byte[] content;
 
     if(!isClasspath(path) && new File(path).exists()) {
-      content = toByteArray(new FileInputStream(path));
-      logger.debug("File loaded from system path : {}", path);
+      try {
+        FileInputStream fis = new FileInputStream(path);
+        content = toByteArray(fis);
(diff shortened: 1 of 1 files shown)


src/main/java/internal/FileHelper.java
```diff
@@ -32,8 +32,15 @@
...
-      logger.debug("File loaded from system path : {}", path);
+      try {
+        FileInputStream fis = new FileInputStream(path);
+        content = toByteArray(fis);
+        fis.close();
+        logger.debug("File loaded from system path : {}", path);
+      }
+      catch (Exception ex) {
+        throw new FileNotFoundException("Unable to load file from path : " + path);
+      }
```
symphony-thibault, line 43: Consider using a [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) statement : 
```java
try(FileInputStream fis = new FileInputStream(path)) {
      content = toByteArray(fis);
      logger.debug("File loaded from system path : {}", path);
    }
    catch (Exception ex) {
      throw new FileNotFoundException("Unable to load file from path : " + path);
    }
```

PR #150, opened 2020-07-01
APP-2781: DatafeedV2 Implementation

symphony-hong:


src/main/java/clients/symphony/api/DatafeedClientV1.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Override
+    public void deleteDatafeed(String id) {
+        throw new UnsupportedOperationException("Delete Datafeed not supported in Datafeed v1");
+    }
+
+    @Override
+    public List<DatafeedEvent> readDatafeed(String id, String ackId) {
+        throw new UnsupportedOperationException("Read Datafeed with ackId not supported in Datafeed v1");
+    }
```
symphony-thibault, line 126: I would actually keep a single `readDatafeed` method but using varargs to define the ackId like : 
```java
public List<DatafeedEvent> readDatafeed(String id, String... ackId) { /*  */ }
```
With this strategy you make your API simpler to implement, to document and to use by developers.

src/main/java/services/AbstractDatafeedEventsService.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.List;
+
+public abstract class AbstractDatafeedEventsService implements DatafeedEventsService {
+
+    protected SymBotClient botClient;
+    protected DatafeedClient datafeedClient;
+    protected List<RoomListener> roomListeners;
+    protected List<IMListener> imListeners;
+    protected List<ConnectionListener> connectionListeners;
+    protected List<ElementsListener> elementsListeners;
```
symphony-thibault, line 20: Every attributes could be `final` right?

src/test/java/it/clients/symphony/api/DatafeedClientV2Test.java
```diff
@@ -0,0 +1,111 @@
...
+                        "}");
+        String datafeedId = datafeedClient.createDatafeed();
+
+        assertEquals("21449143d35a86461e254d28697214b4_f", datafeedId);
+    }
+
+    @Test
+    public void readDatafeedEventSuccess() {
+        stubPost(AgentConstants.READDATAFEEDV2.replace("{id}", "21449143d35a86461e254d28697214b4_f"),
+                "{\n" +
```
symphony-thibault, line 63: I'd prefer to see those payloads in proper `.json` files as part of the test resources

src/main/java/clients/SymBotClient.java
```diff
@@ -268,7 +260,12 @@ public long getBotUserId() {
 
     public DatafeedClient getDatafeedClient() {
         if (datafeedClient == null) {
-            datafeedClient = new DatafeedClient(this);
+            String datafeedVersion = this.config.getDatafeedVersion();
+            if (datafeedVersion != null && datafeedVersion.equalsIgnoreCase("v2")) {
```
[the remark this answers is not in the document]
symphony-elias: And checking the equality the other way removes the need to check for null: `"v2".equalsIgnoreCase(datafeedVersion)`

src/main/java/clients/symphony/api/DatafeedClientV2.java
```diff
@@ -0,0 +1,145 @@
...
+        this.config = client.getConfig();
+    }
+
+    @Override
+    public String createDatafeed() throws SymClientException {
+        Invocation.Builder builder = botClient.getAgentClient()
+                .target(config.getAgentUrl())
+                .path(AgentConstants.CREATEDATAFEEDV2)
+                .request(MediaType.APPLICATION_JSON)
+                .header("sessionToken", botClient.getSymAuth().getSessionToken())
```
symphony-elias, line 41: Maybe `sessionToken` and `keyManagerToken` could be put in AgentConstants.

src/main/java/clients/symphony/api/DatafeedClientV2.java
```diff
@@ -0,0 +1,145 @@
...
+                .header("keyManagerToken", botClient.getSymAuth().getKmToken());
+
+        logger.info("Create datafeed for bot {}..", botClient.getBotUsername());
+        try (Response response = builder.post(null)) {
+            if (response.getStatusInfo().getFamily() != SUCCESSFUL) {
+                try {
+                    handleError(response, botClient);
```
symphony-elias, line 48: My 2 cents: I don't like that logic because it is complex and it hides the fact that `handleError` authenticates again if we have an unauthorized response. And logic is duplicated in `readDatafeed` method.
We will address that later in the v2, I think.

symphony-hong: Yes indeed! Because now I want to do it quickly and make it align with other client, I reused the legacy retry strategy in v1. But clearly we need to improve it in v2!

src/main/java/clients/SymBotClient.java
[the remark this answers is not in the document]
symphony-hong: you mean something like this? 
```
enum Version {
        V1,
        V2;
        
        public Version of(String version) {
            if ("v2".equalsIgnoreCase(version)) {
                return V2;
            } else {
                return V1;
            }
        }
}
```

src/main/java/services/DatafeedEventsServiceV1.java
```diff
@@ -0,0 +1,182 @@
...
+    private ExecutorService pool;
+    private AtomicBoolean stop = new AtomicBoolean();
+    private static int THREADPOOL_SIZE;
+    private static int TIMEOUT_NO_OF_SECS;
+
+    public DatafeedEventsServiceV1(SymBotClient client) {
+        super(client);
+
+        int threadPoolSize = client.getConfig().getDatafeedEventsThreadpoolSize();
+        THREADPOOL_SIZE = threadPoolSize > 0 ? threadPoolSize : 5;
```
symphony-elias, line 37: I would be in favor of putting THREADPOOL_SIZE as a non-static field since it is modified in a non-static method.

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -0,0 +1,89 @@
...
+            } else {
+                //Each bot should subscribe only one datafeed
+                datafeedId = datafeedIds.get(0).getId();
+            }
+
+            //Read datafeed in loop
+            started.set(true);
+            List<DatafeedEvent> events = null;
+            do {
+                events = datafeedClient.readDatafeed(datafeedId);
```
symphony-elias, line 40: Shouldn't we handle the ackId ? First time : no ackId but we should keep the last one and send it back the next time.

symphony-hong: I didn't clearly understand about the ackId. As I understand, when we send back the ackId, all events in this datafeed will be deleted. Do we really need this action?

src/main/java/services/DatafeedEventsServiceV1.java
symphony-hong: As suggestion from @symphony-thibault , we won't change the legacy code of v1, so I only copy-paste code from legacy code for this class. But your comment is right, I also have no idea why THREADPOOL_SIZE is static here!

src/main/java/clients/symphony/api/DatafeedClientV2.java
symphony-thibault: You are right guys, let's keep it as it is in V1 and make error handling way more better in V2 👍

src/main/java/services/DatafeedEventsServiceV1.java
symphony-thibault: Agreed, I want to avoid side effects or breaking changes in the V1 implementation.

src/main/java/services/DatafeedEventsServiceV2.java
symphony-thibault: Digging from this section : https://perzoinc.atlassian.net/wiki/spaces/AP/pages/392593848/1.+Requirements you might find some more documentation/explanation about `ackId`

symphony-thibault: I believe that this field is related to the AWS/SQS acknowledge mode documented [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/getting-started.html#using-client-acknowledge-mode).

src/main/java/clients/symphony/api/DatafeedClient.java
```diff
@@ -1,109 +1,50 @@
...
+    String createDatafeed();
+
+    /**
+     * Read all the events in datafeed.
+     *
+     * @param   id      id of the datafeed to be read
+     * @param   ackId   ackId given to delete all the events in datafeed
+     * @return          list of events in the datafeed
+     */
+    List<DatafeedEvent> readDatafeed(String id, String... ackId);
```
symphony-youri, line 28: aren't we breaking the compatibility here by changing it to an interface? (and by changing some of the methods)

symphony-hong: I don't understand your comment. As I thought, using interface help to avoid breaking the compatibility, isn't it?

symphony-hong:
> I would recommend to put a description body in the git commit logs so we get a bit more context about the change, why it has been done and the technical choices made
> 
> And also to reference the corresponding ticket (it seems we do it even if the repo is open, the tracker is internal) (usually in the commit title SDK-XXX: blahblah) and in the PR branch name

src/main/java/clients/symphony/api/DatafeedClient.java
symphony-youri: For instance let's say that a user of the SDK (an already released version) was overloading the class, now that it is an interface his code does not compile anymore when he updates the SDK. As a SDK or library user I'm not very happy if they make incompatible changes :)
But maybe we don't mind about compatibility, I'm not sure what is the current strategy here.

symphony-hong: I never noticed this example before. The idea behind the interface is keep two version of datafeed is aligned, and it break the compatibility in your example.
Maybe, we can schedule a quick call with @symphony-elias to see how we can modify the strategy here?

symphony-youri: Correction the class is final so my example is not really valid here.

There is the constructor that was public, so by replacing it with an interface we break source compatibility here, although using the constructor was probably not an intended usage of the API (I guess we expect users to use clients.SymBotClient#getDatafeedClient)

I'm not sure if replacing a class with an interface breaks binary compatibility but perhaps we don't guarantee it

Yes we can talk it further but it would be interesting the opinion of someone that worked on the SDK before and know the strategy about compatibility (@symphonydarlys in the absence of @symphony-thibault ?)

src/main/java/clients/symphony/api/constants/DatafeedVersion.java
```diff
@@ -0,0 +1,13 @@
+package clients.symphony.api.constants;
+
+public enum DatafeedVersion {
```
symphony-youri, line 3: do we need to expose this enum publicly in the SDK?

symphony-youri:
Were you able to test it against a bot generated by the bdk?

src/main/java/clients/symphony/api/constants/DatafeedVersion.java
symphony-hong: I expose this enum to use it in classes from other package, ex: DatafeedClient, DatafeedEventsService

symphony-hong:
Read datafeed method should be handled in a new thread instead running on main thread which will block other waiting tasks forever.

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -29,33 +29,34 @@ public DatafeedEventsServiceV2(SymBotClient client) {
...
-            this.started.set(false);
-            logger.error("Something went wrong while reading datafeed", e);
-            logger.info("Sleeping for {} seconds before retrying..", botClient.getConfig().getDatafeedEventsErrorTimeout());
-            sleep(botClient.getConfig().getDatafeedEventsErrorTimeout());
-            readDatafeed();
-        }
 
+                //Read datafeed in loop
+                started.set(true);
+                List<DatafeedEvent> events = null;
```
symphony-youri, line 45: the declaration can be moved inside the do/while

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -29,33 +29,34 @@ public DatafeedEventsServiceV2(SymBotClient client) {
...
+                    if (events != null && !events.isEmpty()) {
+                        handleEvents(events);
+                    }
+                } while (started.get());
+            } catch (Exception e) {
+                this.started.set(false);
+                logger.error("Something went wrong while reading datafeed", e);
+                logger.info("Sleeping for {} seconds before retrying..", botClient.getConfig().getDatafeedEventsErrorTimeout());
+                sleep(botClient.getConfig().getDatafeedEventsErrorTimeout());
+                readDatafeed();
```
symphony-youri, line 57: aren't we going to spawn a thread from a thread each time there is an exception?

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -29,33 +29,34 @@ public DatafeedEventsServiceV2(SymBotClient client) {
...
-            logger.info("Sleeping for {} seconds before retrying..", botClient.getConfig().getDatafeedEventsErrorTimeout());
-            sleep(botClient.getConfig().getDatafeedEventsErrorTimeout());
-            readDatafeed();
-        }
 
+                //Read datafeed in loop
+                started.set(true);
+                List<DatafeedEvent> events = null;
+                do {
+                    events = datafeedClient.readDatafeed(datafeedId, datafeedClient.getAckId());
```
symphony-youri, line 47: is readDatafeed blocking? if so stopDatafeedService is not going to stop the thread until readDatafeed returns

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -0,0 +1,102 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * DatafeedEventsServiceV2 providing services to bot for subscribing the datafeed version 2.
+ */
+class DatafeedEventsServiceV2 extends AbstractDatafeedEventsService {
```
symphony-youri, line 16: for compatibility I would suggest the same approach as for the client, that is to keep the DatafeedEventsService class and put inside it 
and how are we going to access the v2 service?
don't we need to change getDatafeedEventsService ?

symphony-hong: Actually, I'm using the same approach as for the client. You can see it here: https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/d8c88e4c3f1dbc445e092269e4259765de31beb2/src/main/java/services/DatafeedEventsService.java

src/main/java/services/DatafeedEventsServiceV2.java
symphony-hong: for example, when you call: `botClient.getDatafeedService().addListener(listeners)` for the first time, the readDatafeed is call in the constructor, then the `addListener` method will be never called.
I really don't like this approach but I want to make it align with the v1.

symphony-hong:
@symphony-youri I really do not know how to write an unittest for a infinite loop which presents in `DatafeedEventsService` class. Do you any suggestions?

symphony-youri:
As we have a way to stop it, we could have a test that runs the service, mock the client and then call the stopDatafeedService method to stop the thread

symphony-hong:
@symphony-youri I just updated my PR. Can you review this PR please?

symphony-youri:
> does it work in case of failure? because to me an error won't fail the test
> 
> why do we need this to start in a thread? can't we just create the service calls some methods and then the service at the end of the test?

yes tried it locally if you change the asserts the tests still pass

symphony-youri:
Here is the kind of test I would write:

```
@Test
    public void readDatafeedTest() {
        DatafeedV2 id = new DatafeedV2();
        id.setId("123");
        when(datafeedClient.listDatafeedId()).thenReturn(Collections.singletonList(id));
        when(datafeedClient.getAckId()).thenReturn("ack_id_string");

        DatafeedEventsServiceV2 service = new DatafeedEventsServiceV2(datafeedClient);
        service.stopDatafeedService();

        verify(datafeedClient, times(1)).listDatafeedId();
// this is in the loop so it will happen multiple times before the thread is stopped
        verify(datafeedClient, atLeastOnce()).readDatafeed("123", "ack_id_string");
    }
```

without inhering the bot test
no need to json files and mock webservices since we can mock the datafeedclient
DatafeedEventsServiceV2 can be changed to avoid the dependency on Symbotclient and only pass the required info (the datafeed client, bot user id, errors timeout)

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -84,10 +83,16 @@ public void stopDatafeedService() {
         this.started.set(false);
         if (threadPool != null) {
             threadPool.shutdown();
-            if (!threadPool.awaitTermination(AWAIT_TERMINATION, TimeUnit.MILLISECONDS)) {
+            try {
+                if (!threadPool.awaitTermination(AWAIT_TERMINATION, TimeUnit.MILLISECONDS)) {
+                    threadPool.shutdownNow();
+                    if (!threadPool.awaitTermination(AWAIT_TERMINATION, TimeUnit.MILLISECONDS))
+                        System.err.println("Pool did not terminate");
```
symphony-youri, line 90: We always want to use a logger, never system err (you don't know where it ends up whereas we configure logger, have levels and write them to specific location)

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -0,0 +1,131 @@
...
+            threadPool.shutdownNow();
+        }
+        threadPool = Executors.newSingleThreadExecutor();
+        started.set(true);
+        threadPool.submit(() -> {
+            do {
+                try {
+                    readEventsFromDatafeed();
+                } catch (APIClientErrorException e) {
+                    //Datafeed was stale
```
symphony-youri, line 47: we can have a unit test for that too

src/main/java/services/DatafeedEventsServiceV2.java
symphony-hong: oops! My bad.

src/main/java/services/DatafeedEventsServiceV2.java
symphony-hong: Added.

src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -0,0 +1,132 @@
...
+                handleEvents(events);
+            }
+
+        } while (started.get());
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @SneakyThrows
```
symphony-youri, line 81: do we still need it?
from lombok's doc _This somewhat contentious ability should be used carefully, of course_
I would probably avoid using it for public APIs and be explicit on which exceptions can be raised so a user knows which error cases it has to handle

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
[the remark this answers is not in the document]
symphony-hong: it's for mocking final class: DatafeedClientV2 for ex.
https://www.baeldung.com/mockito-final

src/test/java/services/DatafeedEventsServiceTest.java
```diff
@@ -80,24 +84,28 @@ public void stopAndReadAgain() {
...
+        AtomicBoolean finished = new AtomicBoolean();
         DatafeedV2 id = new DatafeedV2();
         id.setId("123");
         when(datafeedClient.listDatafeedId()).thenReturn(Collections.singletonList(id))
                 .thenReturn(Collections.emptyList());
         when(datafeedClient.createDatafeed()).thenReturn("1234");
         when(datafeedClient.readDatafeed("123", "ack_id_string"))
-                .thenThrow(new APIClientErrorException("Feed was deleted because it was a stale feed"))
+                .thenThrow(new APIClientErrorException("Feed was deleted because it was a stale feed"));
+        when(datafeedClient.readDatafeed("1234", "ack_id_string"))
```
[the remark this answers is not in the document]
symphony-hong: because I think it was called with different params in 2 times

src/test/java/services/DatafeedEventsServiceTest.java
```diff
@@ -80,24 +84,28 @@ public void stopAndReadAgain() {
...
-        //Sleep to make sure all steps in process are invoked
-        Thread.sleep(3000);
+        datafeedEventsServiceV2 = new DatafeedEventsServiceV2(symBotClient);
+        DatafeedEventsServiceV2 mockDFService = PowerMockito.spy(datafeedEventsServiceV2);
+        while (!finished.get());
         verify(datafeedClient, times(2)).listDatafeedId();
         verify(datafeedClient, times(1)).createDatafeed();
         verify(datafeedClient, times(1)).readDatafeed("123", "ack_id_string");
         verify(datafeedClient, atLeast(1)).readDatafeed("1234", "ack_id_string");
+        verifyPrivate(mockDFService, times(0)).invoke("sleep", anyInt());
```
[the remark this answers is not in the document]
symphony-hong: but in that way, we will have to public Sleeper and the test constructor to user of sdk. Do we really want that behaviour?

symphony-youri:
Thanks for the rebase, regarding the commit's description we are more interested in the why rather than the what, for instance we will see with the diffs that unit test were added or that DatafeedClient is now final but we are more interested why we did that (for instance to prevent users of the API to subclass) 
This blog post is interesting to read https://chris.beams.io/posts/git-commit/

src/test/java/services/DatafeedEventsServiceTest.java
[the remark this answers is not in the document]
symphony-youri: we can probably make it package protected to not expose it (also in another places in the client I found that https://github.com/apiguardian-team/apiguardian was used to document that classes are internet or not)

the constructor could be package protected as well

src/test/java/services/DatafeedEventsServiceTest.java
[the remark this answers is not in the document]
symphony-youri: ah sorry, missed the 123/1234 difference

src/main/java/clients/symphony/api/datafeeds/DatafeedClientV1.java
```diff
@@ -106,4 +119,30 @@ private void writeDatafeedIdToDisk(SymConfig config, String datafeedId) {
             return datafeedEvents;
         }
     }
+
+
+    /**
+     * {@inheritDoc}
```
symphony-youri, line 125: Might be worth documenting in the Javadoc that those methods are not supported for DFv1

src/main/java/model/datafeed/DatafeedV2.java
```diff
@@ -0,0 +1,26 @@
+package model.datafeed;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties()
```
symphony-youri, line 5: is the annotation used if there are no parameters?

src/main/java/model/datafeed/DatafeedV2.java
```diff
@@ -0,0 +1,26 @@
...
+public class DatafeedV2 {
```
symphony-youri, line 6: We could start using lombok for getter/setter (it seems it started to get used in AgentInfo for instance)

src/test/java/services/DatafeedEventsServiceTest.java
```diff
@@ -0,0 +1,109 @@
...
+        when(symBotClient.getConfig()).thenReturn(config);
+
+        datafeedClient = mock(DatafeedClient.class);
+        when(symBotClient.getDatafeedClient()).thenReturn(datafeedClient);
+        when(datafeedClient.getAckId()).thenReturn("ack_id_string");
+
+    }
+
+    @Test
+    public void readDatafeedTest() {
```
symphony-youri, line 37: Vertical spacing can help to read tests, for instance by adding a bank line between the Given / When / Then, i.e the setup phase, what is tested and the verifications

src/main/java/model/datafeed/DatafeedV2.java
symphony-hong: yes

src/main/java/clients/symphony/api/datafeeds/DatafeedClientV1.java
symphony-hong: I updated the javadoc.

src/test/java/services/DatafeedEventsServiceTest.java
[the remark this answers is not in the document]
symphony-hong: I added the Sleeper and removed Powermock usage from unittest.

PR #174, opened 2020-08-05
Add Configuration features for BDK 2.0

symphony-hong:
Define new structure for Configuration in 2.0 in BdkConfig.java.

Both of yaml and json formats are supported in 2.0, so Yaml parser
is added for parsing Yaml config

LegacyConfig model for supporting the legacy version of configuration.

Configuration is loaded by BdkConfigLoader in three ways:
- From path of config file: BdkConfigLoader#loadFromFile
- From inputstream: BdkConfigLoader#loadFromInputStream
- From classpath: BdkConfigLoader#loadFromClasspath

Set minimum code coverage score: 0.9

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
@@ -1,20 +1,22 @@
 package com.symphony.bdk.core.config;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.json.JsonMapper;
-import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
 import com.symphony.bdk.core.config.legacy.model.LegacySymConfig;
 import com.symphony.bdk.core.config.model.BdkConfig;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.File;
-import java.io.IOException;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.InputStream;
 
 @Slf4j
 public class BdkConfigLoader {
 
-    private static final JsonMapper JSON_MAPPER = new JsonMapper();
-    private static final YAMLMapper YAML_MAPPER = new YAMLMapper();
+    private static final ObjectMapper JSON_MAPPER = new JsonMapper();
 
     /**
      * Load BdkConfig from a file path
@@ -23,20 +25,14 @@ public class BdkConfigLoader {
      *
      * @return Symphony Bot Configuration
      */
-    public static BdkConfig loadFromFile(String configPath) {
+    public static BdkConfig loadFromFile(String configPath) throws JsonProcessingException {
         try {
-            return JSON_MAPPER.readValue(new File(configPath), BdkConfig.class);
-        } catch (IOException e) {
-            log.debug("Config is not in JSON format");
+            File file = new File(configPath);
+            InputStream inputStream = new FileInputStream(file);
+            return loadFromInputStream(inputStream);
+        } catch (FileNotFoundException e) {
+            log.error("Config file is not found.");
         }
-
-        try {
-            return YAML_MAPPER.readValue(new File(configPath), BdkConfig.class);
-        } catch (IOException e) {
-            log.debug("Config is not in Yaml format");
-        }
-
-        log.error("Config is not in a valid format");
         return null;
     }
 
@@ -47,20 +43,18 @@ public class BdkConfigLoader {
      *
      * @return Symphony Bot Configuration
      */
-    public static BdkConfig loadFromInputStream(InputStream inputStream) {
-        try {
-            return JSON_MAPPER.readValue(inputStream, BdkConfig.class);
-        } catch (IOException e) {
-            log.debug("Config is not in JSON format");
+    public static BdkConfig loadFromInputStream(InputStream inputStream) throws JsonProcessingException {
+        if (inputStream != null) {
+            JsonNode jsonNode = BdkConfigParser.parse(inputStream);
+            if (jsonNode != null) {
+                if (jsonNode.at("botUsername") != null) {
+                    LegacySymConfig legacySymConfig = JSON_MAPPER.treeToValue(jsonNode, LegacySymConfig.class);
+                    return BdkConfig.fromLegacyConfig(legacySymConfig);
+                } else {
+                    return JSON_MAPPER.treeToValue(jsonNode, BdkConfig.class);
+                }
+            }
         }
-
-        try {
-            return YAML_MAPPER.readValue(inputStream, BdkConfig.class);
-        } catch (IOException e) {
-            log.debug("Config is not in Yaml format");
-        }
-
-        log.error("Config is not in a valid format");
         return null;
     }
 
@@ -71,86 +65,12 @@ public class BdkConfigLoader {
      *
      * @return Symphony Bot Configuration
      */
-    public static BdkConfig loadFromClasspath(String configPath) {
-        if (!configPath.startsWith(File.separator)) {
-            configPath = File.separator + configPath;
+    public static BdkConfig loadFromClasspath(String configPath) throws JsonProcessingException {
+        InputStream inputStream = BdkConfigLoader.class.getResourceAsStream(configPath);
+        if (inputStream != null) {
+            return loadFromInputStream(inputStream);
         }
-        String externalUrlPath = System.getProperty("user.dir") + configPath;
-
-        BdkConfig config;
-        if ((new File(externalUrlPath)).exists()) {
-            config = loadFromFile(externalUrlPath);
-        } else {
-            InputStream is = BdkConfigLoader.class.getResourceAsStream(configPath);
-            if (is == null) {
-                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(configPath);
-            }
-            config = loadFromInputStream(is);
-        }
-
-        return config;
-    }
-
-    /**
-     * Load LegacySymConfig from a config file
-     *
-     * @param configPath Path to config file
-     *
-     * @return Symphony Bot Legacy Configuration
-     */
-    public static LegacySymConfig loadLegacyConfigFromFile(String configPath) {
-        try {
-            return JSON_MAPPER.readValue(new File(configPath), LegacySymConfig.class);
-        } catch (IOException e) {
-            log.debug("Config is not in JSON format");
-        }
-
-        log.error("Config is not in a valid format");
+        log.error("Config file is not found");
         return null;
-    }
-
-    /**
-     * Load LegacySymConfig from an input stream
-     *
-     * @param inputStream InputStream
-     *
-     * @return Symphony Bot Legacy Configuration
-     */
-    public static LegacySymConfig loadLegacyConfigFromInputStream(InputStream inputStream) {
-        try {
-            return JSON_MAPPER.readValue(inputStream, LegacySymConfig.class);
-        } catch (IOException e) {
-            log.debug("Config is not in JSON format");
-        }
-
-        log.error("Config is not in a valid format");
-        return null;
-    }
-
-    /**
-     * Load LegacySymConfig from a classpath
-     *
-     * @param configPath Classpath to config file
-     *
-     * @return Symphony Bot Legacy Configuration
-     */
-    public static LegacySymConfig loadLegacyConfigFromClasspath(String configPath) {
-        if (!configPath.startsWith(File.separator)) {
-            configPath = File.separator + configPath;
-        }
-        String externalUrlPath = System.getProperty("user.dir") + configPath;
-
-        LegacySymConfig config;
-        if ((new File(externalUrlPath)).exists()) {
-            config = loadLegacyConfigFromFile(externalUrlPath);
-        } else {
-            InputStream is = BdkConfigLoader.class.getResourceAsStream(configPath);
-            if (is == null) {
-                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(configPath);
-            }
-            config = loadLegacyConfigFromInputStream(is);
-        }
-
-        return config;
     }
 }
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
@@ -1,11 +1,13 @@
 package com.symphony.bdk.core.config;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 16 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Load BdkConfig from an InputStream
+     *
+     * @param inputStream InputStream
+     *
+     * @return Symphony Bot Configuration
+     */
+    public static BdkConfig loadFromInputStream(InputStream inputStream) {
+        try {
```
symphony-youri, line 51: this could be factorized with the code above (calling loadFromInputStream from loadFromFile

also perhaps we could try to detect the format based on the extension, using the exceptions feels a bit ugly

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Load BdkConfig from a classpath
+     *
+     * @param configPath Classpath to config file
+     *
+     * @return Symphony Bot Configuration
+     */
+    public static BdkConfig loadFromClasspath(String configPath) {
```
symphony-youri, line 74: isn't redundant with load from stream?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param configPath Path to config file
+     *
+     * @return Symphony Bot Legacy Configuration
+     */
+    public static LegacySymConfig loadLegacyConfigFromFile(String configPath) {
+        try {
+            return JSON_MAPPER.readValue(new File(configPath), LegacySymConfig.class);
+        } catch (IOException e) {
+            log.debug("Config is not in JSON format");
+        }
```
symphony-youri, line 106: What about versioning the config file (with a version number in it) and have the same config object (wrapping the legacysymconfig in a bdkconfig)?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            return JSON_MAPPER.readValue(inputStream, BdkConfig.class);
+        } catch (IOException e) {
+            log.debug("Config is not in JSON format");
+        }
+
+        try {
+            return YAML_MAPPER.readValue(inputStream, BdkConfig.class);
+        } catch (IOException e) {
+            log.debug("Config is not in Yaml format");
```
symphony-youri, line 60: Yaml -> YAML

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        log.error("Config is not in a valid format");
```
symphony-youri, line 63: I would add more details about the supported formats (just saying that JSON and YAML are supported)

symphony-youri: Also what if the file path does not exist? we do not distinguish between file not found and wrong format

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkSSLConfig.java
```diff
@@ -0,0 +1,14 @@
+package com.symphony.bdk.core.config.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class BdkSSLConfig {
```
symphony-youri, line 10: I would keep acronyms as camel case to avoid wondering witch case to use (for if you have an HTTPSSL class it would be weird, with HttpSsl we are consistent no matter what)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkBotConfig.java
```diff
@@ -0,0 +1,16 @@
...
+
+@Getter
+@Setter
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class BdkBotConfig {
+
+    private String username;
+    private String privateKeyPath;
+    private String certificatePath;
+    private String getCertificatePassword;
```
symphony-youri, line 15: get?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkSSLConfig.java
```diff
@@ -0,0 +1,14 @@
...
+
+    private String truststorePath;
```
symphony-youri, line 12: trustStorePath maybe?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
```diff
@@ -0,0 +1,33 @@
+package com.symphony.bdk.core.config.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@JsonIgnoreProperties(ignoreUnknown = true)
```
symphony-youri, line 9: that could be configured globally in the mapper to avoid repeating it

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
```diff
@@ -0,0 +1,33 @@
...
+public class BdkClientConfig {
+
+    private static final String DEFAULT_SCHEME = "https";
+    private static final int DEFAULT_HTTPS_PORT = 443;
+
+    private String scheme = DEFAULT_SCHEME;
+    private String url;
```
symphony-youri, line 16: does this mean we have to configure it like:

agent:
 url:
 port:
 scheme:
... ? it feels a bit cumbersome no?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
symphony-hong: I also have an idea to detect the format by using regular expression but I'm not sure if it's a good way or not. I searched from internet and they all suggested using exceptions.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
symphony-hong: > I would add more details about the supported formats (just saying that JSON and YAML are supported)

I think it will be determined in the documentation

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
symphony-hong: For example, we will have:
```
agent:
   url:
   port:
   scheme:
   context:
pod:
   url:
   port:
   scheme:
   context:
```
instead of in v1:
```
  agentHost:
  agentPort:
  agentContextPath:
  podHost:
  podPort:
  podContextPath:
```
I think it would be clearer and more consistent

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
symphony-thibault: Correct @symphony-youri, even when loading a legacy config file it should return a `BdkConfig` object. As well, I'd prefer to not have specific methods for legacy config loading, we should be able to detect it before mapping to target class : 

```
JsonNode config = parseConfigFile(path or inputstream or whatever)
if (config.at("/botUsername") != null) {
    // process legacy config format
} else {
    // new config format
}
```

As I also mentioned yesterday, we could also adopt the same strategy as used by `docker-compose.yml` file (cf. https://docs.docker.com/compose/) : 
```yaml
version: '2.0'
services:
  web:
    build: .
    ports:
    - "5000:5000"
    volumes:
    - .:/code
    - logvolume01:/var/log
    links:
    - redis
  redis:
    image: redis
volumes:
  logvolume01: {}
```
Where the version of the config format is part of the config file

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/legacy/model/LegacySymConfig.java
```diff
@@ -0,0 +1,154 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.io.File;
+import java.util.ArrayList;
+
+@Getter
+@Setter
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class LegacySymConfig {
```
symphony-thibault, line 13: Mark this class as deprecated : 

```java
@Getter
@Setter
@JsonIgnoreProperties(ignoreUnknown = true)
@Deprecated
@API(status = API.Status.DEPRECATED)
public class LegacySymConfig {

}
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/legacy/model/RetryConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.ToString;
+import org.apiguardian.api.API;
+
+/**
+ * Sub-configuration class for Retry parametrization. Experimental, the contract might change in the future.
+ */
+@ToString
+@Getter @Setter
+@API(status = API.Status.EXPERIMENTAL)
+public class RetryConfiguration {
```
symphony-thibault, line 14: Mark as deprecated and rename to `LegacyRetryConfiguration `: 

```java
@ToString
@Getter @Setter
@Deprecated
@API(status = API.Status.DEPRECATED)
public class LegacyRetryConfiguration {

}
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkSSLConfig.java
symphony-hong: in V1, it was defined as "truststorePath". And I also think "truststore" is a complete word.
https://docs.oracle.com/cd/E19830-01/819-4712/ablqw/index.html

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
symphony-thibault, line 16: `host` sounds better than `url` WDYT? 

@symphony-youri actually most the time developers will only have to configure `host` and `port` for on-prem components, for on-cloud components it's even simpler as they would only have to configure the `host` (e.g. `devx1.symphony.com` or `corporate.symphony.com`)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
```diff
@@ -0,0 +1,33 @@
...
+    private int port;
+    private String context = "";
+
+    public String getBasePath() {
+        return this.scheme + "://" + this.url + ":" + this.port;
+    }
```
symphony-thibault, line 22: Context is not part of the `basePath`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkSSLConfig.java
symphony-thibault: +1 : `trustStorePath` and `trustStorePassword` (cf. Oracle doc : https://docs.oracle.com/cd/E19509-01/820-3503/6nf1il6er/index.html)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
```diff
@@ -0,0 +1,33 @@
...
+        } catch (IOException e) {
+            log.debug("Config file is not in JSON format");
+        }
+
+        try {
+            return YAML_MAPPER.readTree(inputStream);
+        } catch (IOException e) {
+            log.debug("Config file is not in YAML format");
+        }
+        log.error("Config file is not in a valid format");
```
symphony-thibault, line 30: You should throw a `BdkConfigException extends Exception` with details message

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private BdkClientConfig agent = new BdkClientConfig();
+    private BdkClientConfig pod = new BdkClientConfig();
+    private BdkClientConfig keyManager = new BdkClientConfig();
+
+    private BdkBotConfig bot = new BdkBotConfig();
+    private BdkExtAppConfig app = new BdkExtAppConfig();
+    private BdkSslConfig ssl = new BdkSslConfig();
+
+    public static BdkConfig fromLegacyConfig(LegacySymConfig legacySymConfig) {
```
symphony-thibault, line 23: Move this method to an mapper class : 
```java
public class LegacyConfigMapper {

     public static BdkConfig map(LegacySymConfig legacySymConfig) {}
}
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkSslConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class BdkSslConfig {
+
+    private String truststorePath;
+    private String truststorePassword;
```
symphony-thibault, line 13: `trustStorePath` and `trustStorePassword `

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
symphony-thibault: @symphony-hong +1

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigLoader.java
symphony-youri: Yes but as we are very much aware developers don't read the doc :) 
If we can provide helpful error message that suggest how to fix an error that will improve the developer's experience

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
symphony-youri: would not it be easier to just configure an URL?

PR #175, opened 2020-08-05
APP-2884

symphony-hong:
InputStream cannot be read again after the first read
-> read content from InputStream first and detect format
from the extracted content

Using JsonMapper#convertValue instead of JsonMapper#readValue
to avoid JsonProcessingException

Move LegacyConfigMapper to legacy package

Using JsonNode#isMissingNode() instead of equals null

Add very first unittest to make sure the Loader is working

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
@@ -8,8 +8,12 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLMapper;
 import com.symphony.bdk.core.exceptions.BdkConfigException;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.stream.Collectors;
 
 @Slf4j
 class BdkConfigParser {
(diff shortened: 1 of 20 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
```diff
@@ -20,14 +24,18 @@
     public static JsonNode parse(InputStream inputStream) throws BdkConfigException {
         JSON_MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
         YAML_MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        String content = new BufferedReader(
```
symphony-youri, line 27: since we have commons io as a dependency we could use IOUtils.toString(inputstream)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigLoaderTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    public void loadFromYamlFileTest() throws JsonProcessingException, BdkConfigException {
+        String configPath = System.getProperty("user.dir") + "/src/test/resources/config.yaml";
+        BdkConfig config = BdkConfigLoader.loadFromFile(configPath);
+        assertEquals(config.getBot().getUsername(), "tibot");
+    }
+
+    @Test
+    public void loadFromJsonFileTest() throws JsonProcessingException, BdkConfigException {
+        String configPath = System.getProperty("user.dir") + "/src/test/resources/config.json";
```
symphony-youri, line 37: Does it work in IDEA? not sure the default current folder is the module's folder

PR #176, opened 2020-08-06
Add unittest and documentations for BDK configuration

symphony-hong:
Add unittest and documentations for BDK configuration

Move BdkConfigException to com.symphony.bdk.core.config.exceptions
to avoid many types of exceptions in a single packages

Add a condition in BdkConfigParser#parse to check if invalid YAML config is parsed to a single
text node.


symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
```diff
@@ -35,7 +35,9 @@ public static JsonNode parse(InputStream inputStream) throws BdkConfigException
         }
 
         try {
-            return YAML_MAPPER.readTree(content);
+            JsonNode jsonNode =  YAML_MAPPER.readTree(content);
```
symphony-youri, line 38: We should probably make BdkConfigParser internal right? 
I see it is already does not have public visibility, maybe annotate it with API guardian?

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigLoaderTest.java
```diff
@@ -1,53 +1,79 @@
...
+        InputStream inputStream = BdkConfigLoaderTest.class.getResourceAsStream("/config/config.json");
         BdkConfig config = BdkConfigLoader.loadFromInputStream(inputStream);
         assertEquals(config.getBot().getUsername(), "tibot");
     }
 
     @Test
-    public void loadFromYamlFileTest() throws JsonProcessingException, BdkConfigException {
-        String configPath = System.getProperty("user.dir") + "/src/test/resources/config.yaml";
+    public void loadFromYamlFileTest() throws BdkConfigException {
+        String configPath = System.getProperty("user.dir") + "/src/test/resources/config/config.yaml";
```
symphony-thibault, line 30: I'm not sure that this is the safest way to locate the test resources. I'd personally use another strategy :
- using https://www.baeldung.com/junit-5-temporary-directory#1-tempdir-as-a-method-parameter
- copy the config file from the classpath to tempdir
- read config from tempdir

WDYT?

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigLoaderTest.java
```diff
@@ -1,53 +1,79 @@
...
+        assertEquals(exception.getMessage(), "Config file is not found");
+    }
+
+    @Test
+    public void loadLegacyFromInputStreamTest() throws BdkConfigException {
+        InputStream inputStream = BdkConfigLoaderTest.class.getResourceAsStream("/config/legacy_config.json");
+        BdkConfig config = BdkConfigLoader.loadFromInputStream(inputStream);
+        assertEquals(config.getBot().getUsername(), "tibot");
+        assertEquals(config.getBot().getPrivateKeyPath(), "/Users/thibault.pensec/local/conf/agent/privatekey.pem");
+        assertEquals(config.getApp().getPrivateKeyPath(), "/Users/thibault.pensec/local/conf/agent/privatekey.pem");
```
symphony-thibault, line 69: Please remove "thibault.pensec" from the test data 😅  I'll do the same with my PR !

symphony-bdk-core/src/test/resources/config/legacy_config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "podHost": "devx1.symphony.com",
+  "agentHost": "devx1.symphony.com",
+  "keyAuthHost": "devx1.symphony.com",
+  "botUsername": "tibot",
+  "botPrivateKeyPath": "/Users/thibault.pensec/local/conf/agent/",
```
symphony-thibault, line 6: Please remove "thibault.pensec" !

symphony-bdk-core/src/test/resources/config/legacy_config.json
```diff
@@ -0,0 +1,11 @@
...
+  "botPrivateKeyName": "privatekey.pem",
+  "appId": "tibapp",
+  "appPrivateKeyPath": "/Users/thibault.pensec/local/conf/agent/",
```
symphony-thibault, line 9: Same...

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigLoaderTest.java
symphony-hong: It's a good idea! I will try that way. 👍

PR #179, opened 2020-08-13
APP-2898: Fix exception handling when reading a faulty datafeed

symphony-hong:
Once reading a datafeed throws an APIClientException, it means that
the datafeed is faulty. Before, this datafeed is automatically removed
after being read but now it's not removed anymore. Therefore, the
faulty datafeed should be deleted manually after the exception is thrown.

### DIFF
--- a/symphony-bdk-legacy/symphony-api-client-java/src/main/java/services/DatafeedEventsServiceV2.java
+++ b/symphony-bdk-legacy/symphony-api-client-java/src/main/java/services/DatafeedEventsServiceV2.java
@@ -51,6 +51,8 @@ class DatafeedEventsServiceV2 extends AbstractDatafeedEventsService {
             do {
                 try {
                     readEventsFromDatafeed();
+                } catch (APIClientErrorException e) {
+                    logger.debug("Something went wrong rather than faulty datafeed", e);
                 } catch (Exception e) {
                     logger.error("Something went wrong while reading datafeed", e);
                     logger.info("Sleeping for {} seconds before retrying..", botClient.getConfig().getDatafeedEventsErrorTimeout());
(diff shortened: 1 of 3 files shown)


symphony-hong:
> Can we add a unit test showing the problem?

symphony-hong:
> Can we add a unit test showing the problem?

There's already an unittest for this case. I just updated the verifying scenarios.

symphony-bdk-legacy/symphony-api-client-java/src/main/java/services/DatafeedEventsServiceV2.java
```diff
@@ -74,10 +71,19 @@ private void readEventsFromDatafeed() {
...
+            try {
+                List<DatafeedEvent> events = datafeedClient.readDatafeed(datafeedId, datafeedClient.getAckId());
+                if (events != null && !events.isEmpty()) {
+                    handleEvents(events);
+                }
+            } catch (APIClientErrorException e) {
+                //Datafeed was stale
+                logger.debug("Something went wrong with this datafeed: {}", datafeedId);
+                logger.info(e.getMessage());
+                datafeedClient.deleteDatafeed(datafeedId);
```
symphony-youri, line 85: will that call always succeed? even if the data feed is faulty or not known anymore? if not don't we risk having a weird chain of logs?

also in   services.DatafeedEventsServiceV2#readDatafeed
we already had
   } catch (APIClientErrorException e) {
                    //Datafeed was stale
                    logger.info(e.getMessage());

does this make sense to have two catches with the same comment? should'nt it be only one?



Looking at that code again I asked my self should we delete the datafeed in stopDatafeedService? or at the end of readEventsFromDatafeed?

symphony-hong: > will that call always succeed? even if the data feed is faulty or not known anymore? if not don't we risk having a weird chain of logs?
This delete call will return 400 error if the the datafeed is not found -> APIClientError, should we add another try/catch inside catch block?

> does this make sense to have two catches with the same comment? should'nt it be only one?
I removed the one in `services.DatafeedEventsServiceV2#readDatafeed` 

>Looking at that code again I asked my self should we delete the datafeed in stopDatafeedService? or at the end of readEventsFromDatafeed?

I'm not sure about that idea, because I think the datafeed is still reusable after the service stops.

symphony-youri: perhaps the 400 error is something we would like to catch and just log in debug if this happens instead of waiting before retrying

PR #180, opened 2020-08-17
APP-2845 : BDK certificate authentication

symphony-elias:
This intends to implement APP-2845.


symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientJersey2.java
```diff
@@ -205,6 +211,23 @@ public ApiClient setBasePath(String basePath) {
     return this;
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public ApiClient setClientCertificate(String certificatePassword, String password) throws GeneralSecurityException, IOException {
```
symphony-youri, line 218: can we keep the client final and have different constructors instead of mutating the object?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionImpl.java
```diff
@@ -14,12 +14,12 @@
...
 public class AuthSessionImpl implements AuthSession {
 
-  private final BotAuthenticatorRsaImpl authenticator;
+  private final BotAuthenticator authenticator;
 
   private String sessionToken;
   private String keyManagerToken;
 
-  public AuthSessionImpl(@Nonnull BotAuthenticatorRsaImpl authenticator) {
+  public AuthSessionImpl(@Nonnull BotAuthenticator authenticator) {
```
symphony-hong, line 22: Not really related to your PR but do you think we need to retrieve session token and keymanager token at the moment that the `AuthSession` is initiated, instead of let it null and retrieve them by calling `refresh()`?

```
public AuthSessionImpl(@Nonnull BotAuthenticatorRsaImpl authenticator) throws AuthUnauthorizedException {
    this.authenticator = authenticator;
    this.sessionToken = authenticator.retrieveSessionToken();
    this.keyManagerToken = authenticator.retrieveKeyManagerToken();
  }
```

symphony-elias: Maybe let us check that with @symphony-thibault

symphony-elias: Already implemented in https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/pull/181

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiClient.java
```diff
@@ -16,18 +15,18 @@
   /**
    * Invoke API by sending HTTP request with the given options.
    *
-   * @param <T> Type
-   * @param path The sub-path of the HTTP URL
-   * @param method The request method, one of "GET", "POST", "PUT", "HEAD" and "DELETE"
-   * @param queryParams The query parameters
-   * @param body The request body object
+   * @param <T>          Type
```
symphony-youri, line 18: Looks like different code styles are used, we should use the same everywhere

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/pom.xml
```diff
@@ -142,6 +142,10 @@
             <artifactId>mockserver-netty</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
symphony-youri, line 147: we can use the provided scope to avoid exposing it (https://projectlombok.org/setup/maven)

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -0,0 +1,171 @@
...
+    basePath = "https://acme.symphony.com";
+    keyStorePath = null;
+    keyStorePassword = null;
+    trustStorePath = null;
+    trustStorePassword = null;
+    defaultHeaders = new HashMap<>();
+    connectionTimeout = 0;
+    readTimeout = 0;
+    temporaryFolderPath = null;
+    userAgent("Symphony BDK/2.0/java");
```
symphony-youri, line 45: Maybe we could try to be close the "standard" for user-agent (https://stackoverflow.com/questions/2601372/what-is-the-standard-format-for-a-browsers-user-agent-string)
So maybe something like _Symphony BDK/2.0 java/1.8.212_ ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
```diff
@@ -12,6 +12,7 @@
     private BdkClientConfig agent = new BdkClientConfig();
     private BdkClientConfig pod = new BdkClientConfig();
     private BdkClientConfig keyManager = new BdkClientConfig();
+    private BdkClientConfig sessionAuth = new BdkClientConfig();
```
symphony-youri, line 15: makes me wonder if as a user of the SDK I really need to understand all the complexity of agent/pod/keyManager/sessionAuth and need to configure all of that

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiClient.java
symphony-thibault: We should use : https://github.com/SymphonyOSF/appcentral/blob/master/.editorconfig

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/BotAuthenticator.java
```diff
@@ -16,4 +18,18 @@
...
+   * @throws AuthUnauthorizedException in case of authentication failure
+   */
+  @NonNull String retrieveSessionToken() throws AuthUnauthorizedException;
+
+  /**
+   *
+   * @return a new key manager token
+   * @throws AuthUnauthorizedException in case of authentication failure
+   */
+  @NonNull String retrieveKeyManagerToken() throws AuthUnauthorizedException;
```
symphony-thibault, line 34: The initial idea to retrieve `sessionToken` and `keyManagerToken` was actually to use the `AuthSession#getSessionToken()` and `AuthSession#getKeyManagerToken()`. Adding those methods here is redundant.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
symphony-thibault: In most of the cases (unfortunately) yes you'll have to configure all of these parameters. We have to keep in mind that most of the customers that are implementing their workflows using bots and apps are based on on-prem architecture along with cert authentication. I don't see how we could avoid configuring those components =/ 

On the other hand, the `yeoman` generator really helps to quickly setup a BDK project with all of these parameters already configured (mostly used during hackathons) when Agent/KM are on-cloud and bots are using RSA authentication.

symphony-hong: this conversation reminds me a message from Pat O'Brien in SDK channel:

> When using rsa auth I want the sdk to ignore sessionauth url and use pod url. sessionauth urtl is for ssl cert auth and it confuses customers

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionImpl.java
symphony-elias: @symphony-thibault What do you think ?

symphony-thibault: Correct, let's follow the same way as already done here : https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/23798f6d6200f7e6fed1dac1d3d7828962f9b7ed/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorRsaImpl.java#L55

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiClientBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * @param key name of the header to be added in all HTTP calls
+   * @param value header value
+   * @return the updated instance of {@link ApiClientBuilder}
+   */
+  ApiClientBuilder withDefaultHeader(String key, String value);
+
+  /**
+   * @param temporaryFolderPath temporary folder path where to download files sent as response
```
symphony-thibault, line 46: Do we know what is the default value?

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -0,0 +1,171 @@
...
+
+  public ApiClientBuilderJersey2() {
+    basePath = "https://acme.symphony.com";
+    keyStoreBytes = null;
+    keyStorePassword = null;
+    trustStoreBytes = null;
+    trustStorePassword = null;
+    defaultHeaders = new HashMap<>();
+    connectionTimeout = 0;
+    readTimeout = 0;
```
symphony-thibault, line 43: Looks risky to set those default values at `0`: if not overridden thought the builder it might create some confusion with weird behaviour. Just found : https://support.chargebee.com/support/discussions/topics/40956 that could help to set those default values.

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientBuilderJersey2.java
symphony-thibault: In the future that would be nice to retrieve the real version from the `pom.xml`, potentially by using Maven resources filtering (cf. https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html). Could you create a story for that?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionCertImpl.java
```diff
@@ -0,0 +1,36 @@
...
+  @Override
+  public void refresh() throws AuthUnauthorizedException {
+    this.sessionToken = authenticator.retrieveSessionToken();
+    this.keyManagerToken = authenticator.retrieveKeyManagerToken();
+  }
+
+  protected BotAuthenticatorCertImpl getAuthenticator() {
+    return authenticator;
+  }
+}
```
symphony-thibault, line 36: Javadoc missing on class and methods, even if `{@inheritDoc}` for methods 😉

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorCertImpl.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apiguardian.api.API;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Bot authenticator certificate implementation.
+ *
+ * @see
+ * <a href="https://developers.symphony.com/symphony-developer/docs/bot-authentication-workflow-1">Bot
+ * Authentication Workflow</a>
```
symphony-thibault, line 22: Weird behaviour if the formatter, it could be written on a single line. Do we know if the Javadoc is still valid even if written like this?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorRsaImpl.java
```diff
@@ -48,20 +50,17 @@ public BotAuthenticatorRsaImpl(
...
-  public AuthSession authenticateBot() throws AuthUnauthorizedException {
-    AuthSession authSession = new AuthSessionImpl(this);
-    authSession.refresh();
-    return authSession;
+  public @Nonnull AuthSession authenticateBot() {
+    return new AuthSessionRsaImpl(this);
   }
 
-  protected String retrieveSessionToken() throws AuthUnauthorizedException {
+  public String retrieveSessionToken() throws AuthUnauthorizedException {
```
symphony-thibault, line 58: Why it has to be public? Class `AuthSessionRsaImpl.java` is part of the same package so it should be fine to keep it as `protected`

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiClientBuilder.java
```diff
@@ -0,0 +1,68 @@
+package com.symphony.bdk.core.api.invoker;
+
+/**
+ * Builder class to create a new instance of an {@link ApiClient}
+ */
+public interface ApiClientBuilder {
```
symphony-thibault, line 6: + annotation `@API(status = API.Status.STABLE)`

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -0,0 +1,171 @@
...
+import javax.ws.rs.client.ClientBuilder;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
+
+/**
+ * Specific implementation of {@link ApiClientBuilder} which creates a new instance of an {@link ApiClientJersey2}.
+ */
+public class ApiClientBuilderJersey2 implements ApiClientBuilder {
```
symphony-thibault, line 23: + annotation  `@API(status = API.Status.STABLE)`

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientBuilderProviderJersey2.java
```diff
@@ -0,0 +1,21 @@
+package com.symphony.bdk.core.api.invoker.jersey2;
+
+import com.symphony.bdk.core.api.invoker.ApiClientBuilder;
+import com.symphony.bdk.core.api.invoker.ApiClientBuilderProvider;
+
+/**
+ * Provides new {@link ApiClientBuilderJersey2} implementation of the {@link ApiClientBuilder} interface.
+ */
+public class ApiClientBuilderProviderJersey2 implements ApiClientBuilderProvider {
```
symphony-thibault, line 9: + annotation `@API(status = API.Status.INTERNAL)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionCertImpl.java
```diff
@@ -0,0 +1,36 @@
+package com.symphony.bdk.core.auth.impl;
+
+import com.symphony.bdk.core.auth.AuthSession;
+import com.symphony.bdk.core.auth.exception.AuthUnauthorizedException;
+
+public class AuthSessionCertImpl implements AuthSession {
```
symphony-thibault, line 6: + annotation `@API(status = API.Status.INTERNAL)`

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiClientBuilder.java
symphony-elias: Value is used here : https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientJersey2.java#L528
If it is not set, it uses value of property "java.io.tmpdir" which depends on the OS : https://dzone.com/articles/working-with-temporary-filesfolders-in-java

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorCertImpl.java
symphony-elias: I guess so but max_line_length is set to 100 in the .editorconfig file which is quite small (IntelliJ default value is 120)

symphony-youri: While I find it nice to automatically wrap long lines I'd say for comments it is up to the writer to make it readable
Often it breaks URLs
I found "Wrap at right margin" that could work for javadoc comments
image

symphony-youri: In .editorconfig if I use 
_ij_java_wrap_comments = false
ij_java_wrap_long_lines = false_

it seems to not wrap comments (but it still wraps code)

symphony-youri: (the 100 margin comes from SBE I think, I agree that we could probably move to 120)

symphony-bdk-core-invokers/symphony-bdk-core-invoker-jersey2/src/main/java/com/symphony/bdk/core/api/invoker/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -0,0 +1,180 @@
...
+    httpClient.property(ClientProperties.READ_TIMEOUT, readTimeout);
+
+    return new ApiClientJersey2(httpClient, basePath, defaultHeaders, temporaryFolderPath);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public ApiClientBuilder basePath(String basePath) {
```
symphonydarlys, line 75: with this great pattern builder we should have all method with the same format name, I mean we should have all method names with the prefix with. withBastPath(String), withUserAgent, ....

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorCertImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    return doRetrieveToken(this.keyAuthClient);
+  }
+
+  private String doRetrieveToken(ApiClient client) throws AuthUnauthorizedException {
+    try {
+      final Token token = new CertificateAuthenticationApi(client).v1AuthenticatePost();
+      log.debug("{} successfully retrieved.", token.getName());
+      return token.getToken();
+    } catch (ApiException ex) {
+      if (ex.getCode() == 401) {
```
symphonydarlys, line 61: for all http error we must use this class provided by java : https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#HTTP_UNAUTHORIZED

symphonydarlys:
replace all http error by https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html values.
Please document all classes and public methods to help developers to understand symphony Java-BDK

symphony-thibault (approved):
LGTM 👍 Very good job !

PR #181, opened 2020-08-19
APP-2904: Datafeed V1 implementation on BDK 2.0

symphony-hong:
Create interface DatafeedService with public abstract methods to expose to users, abstract class AbstractDatafeedService implementing DatafeedService, define protected methods to be used in both DatafeedService V1 and V2, two classes DatafeedServiceV1 and DatafeedServiceV2 has their own implementation for DF v1 and v2.

Create BotInfoService to retrieve information about the bot itself to be used in others services.

Modify the AuthSessionImpl, retrieve the session token and keymanager token at the moment the AuthSessionImpl instance was created, instead of let it null and retrieve by calling AuthSessionImpl#refresh.

Create a function in ApiClientFactory#getPodClient to get the pod client to make a call to pod.

Create ApiException#isClientError and ApiException#isServerError to check if an ApiException is client error or server error

Retry Strategy, BdkRetryConfig was created including maxAttempts, initialIntervalMillis, maxIntervalMillis, multiplier. Retry mechanism was implemented by using resilience4j-retry.

An example using datafeed v1 service.

Unittest and documentation is added.


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedService.java
```diff
@@ -0,0 +1,30 @@
+package com.symphony.bdk.core.service.datafeed;
+
+public interface DatafeedService {
+
+    /**
+     * Start the datafeed events service
+     *
+     */
+    void start() throws Throwable;
```
symphony-elias, line 9: I don't really like `throws Throwable`, this is too generic in my opinion.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,148 @@
...
+            try {
+                List<V4Event> events = datafeedApi.v4DatafeedIdReadGet(datafeedId, authSession.getSessionToken(), authSession.getKeyManagerToken(), null);
+                if (events != null && !events.isEmpty()) {
+                    handleV4EventList(events);
+                }
+            } catch (ApiException e) {
+                log.error("Error {}: {}", e.getCode(), e.getMessage());
+                if (e.isUnauthorized()) {
+                    log.info("Re-authenticate and try again");
+                    try {
```
symphony-elias, line 74: Do we need a try catch just to log and throw the very same exception ? Exception with stack trace will be shown in the logs, won't it?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,148 @@
...
+        return this.retry.executeCheckedSupplier(() -> {
+            try {
+                String id = this.datafeedApi.v4DatafeedCreatePost(authSession.getSessionToken(), authSession.getKeyManagerToken()).getId();
+                this.writeDatafeedIdToDisk(id);
+                log.debug("Datafeed: {} was created and saved to disk", id);
+                return id;
+            } catch (ApiException e) {
+                log.error("Error {}: {}", e.getCode(), e.getMessage());
+                if (e.isUnauthorized()) {
+                    try {
```
symphony-elias, line 102: Same comment as above

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,148 @@
...
+    private String retrieveDatafeedIdFromDisk() {
+        String datafeedId = null;
+        try {
+            File file = this.getDatafeedIdFile();
+            Path datafeedIdPath = Paths.get(file.getPath());
+            String[] persistedDatafeed = Files.readAllLines(datafeedIdPath).get(0).split("@");
+            datafeedId = persistedDatafeed[0];
+            log.info("Retrieve datafeed id from persisted file: {}", datafeedId);
+        } catch (IOException e) {
+            log.info("No datafeed id is persisted");
```
symphony-elias, line 123: IMHO this log is quite confusing, maybe we should say something like : `No persisted datafeed id could be retrieved from the filesystem`.
Maybe we would like to log the exception message `e.getMessage()` (which could be file not found, for instance)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedService.java
symphony-hong: Because inside implementation of `start()` in V1 or even in V2 in future, I use resilience4j-retry to implement the retry mechanism and `Retry#executeCheckedSupplier` throws `Throwable`. So I'm not sure if we have other choices.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/BdkExponentialFunction.java
```diff
@@ -0,0 +1,28 @@
+package com.symphony.bdk.core.util;
+
+import com.symphony.bdk.core.config.model.BdkRetryConfig;
+import io.github.resilience4j.core.IntervalFunction;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class BdkExponentialFunction {
+
+    public static IntervalFunction ofExponentialBackoff(BdkRetryConfig retryConfig) {
```
symphony-youri, line 10: it seems resilience4j is exposing a function already: https://resilience4j.readme.io/docs/retry#use-a-custom-intervalfunction
do we do something different here?

looks like the getMaxIntervalMillis is what we do differently

symphony-youri: but in a sense since we have maxAttempts(bdkRetryConfig.getMaxAttempts())
don't we already limit the max interval duration?

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/DatafeedExampleMain.java
```diff
@@ -0,0 +1,128 @@
...
+
+            @Override
+            public void onSymphonyElementsAction(V4Event event) {
+
+            }
+        });
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        executorService.submit(() -> {
+            try {
+                ds.start();
```
symphony-youri, line 121: or we can just ds.start() in the main thread (and kill the bot manually when needed)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRetryConfig.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Setter;
+
+@Getter
+@Setter
+public class BdkRetryConfig {
+
+    public static final int DEFAULT_MAX_ATTEMPTS = 10;
+    public static final long DEFAULT_INITIAL_INTERVAL_MILLIS = 500L;
+    public static final double DEFAULT_MULTIPLIER = 2;
+    public static final long DEFAULT_MAX_INTERVAL_MILLIS = 5*60*1000;
```
symphony-youri, line 13: with those defaults it looks like (if I understand resilience4j correctly) that the retries are going to be 
500, 1000, 2000, 4000, 8000, 16000, 32000, 64000, 128000, 360000 and that's the 10th attempts so we stop

so do we need a 360000 max interval limit in the end? (see also the comment below on the Exponential interval function)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
```diff
@@ -17,4 +17,7 @@
     private BdkExtAppConfig app = new BdkExtAppConfig();
     private BdkSslConfig ssl = new BdkSslConfig();
 
+    private BdkRetryConfig retry = new BdkRetryConfig();
```
symphony-youri, line 20: we have it at the same level here and also part of BdkDatafeedConfig, is it normal?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Setter;
+
+import java.io.File;
+
+@Getter
+@Setter
+public class BdkDatafeedConfig {
+
+    private String version = "v1";
+    private String datafeedIdFilePath;
```
symphony-youri, line 13: as we are in the datafeed path of the configuration should we add datafeed again?
config:
  datafeed:
   idFilePath: ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
symphony-youri, line 20: is the retry configuration something specific to DF API or should it apply to all clients?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedService.java
symphony-youri: We could wrap those exceptions
I agree that is feels a bit too generic and it does not tell the user which errors could happen
at least the IllegalStateException if you start it twice

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+        log.debug("Start retrieving datafeed id from disk");
+        String datafeedId = null;
+        try {
+            File file = this.getDatafeedIdFile();
+            Path datafeedIdPath = Paths.get(file.getPath());
+            String[] persistedDatafeed = Files.readAllLines(datafeedIdPath).get(0).split("@");
+            datafeedId = persistedDatafeed[0];
+            log.info("Retrieve datafeed id from persisted file: {}", datafeedId);
+        } catch (IOException e) {
+            log.debug(e.getMessage());
```
symphony-youri, line 112: as it is expected does it make sense to log the exception?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private String retrieveDatafeedIdFromDisk() {
+        log.debug("Start retrieving datafeed id from disk");
+        String datafeedId = null;
+        try {
+            File file = this.getDatafeedIdFile();
+            Path datafeedIdPath = Paths.get(file.getPath());
+            String[] persistedDatafeed = Files.readAllLines(datafeedIdPath).get(0).split("@");
+            datafeedId = persistedDatafeed[0];
```
symphony-youri, line 109: what is the file is empty or has been changed manually?
We should probably just ignore it silently (here it will break with an exception)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+        return this.retry.executeCheckedSupplier(() -> {
+            try {
+                String id = this.datafeedApi.v4DatafeedCreatePost(authSession.getSessionToken(), authSession.getKeyManagerToken()).getId();
+                this.writeDatafeedIdToDisk(id);
+                log.debug("Datafeed: {} was created and saved to disk", id);
+                return id;
+            } catch (ApiException e) {
+                log.error("Error {}: {}", e.getCode(), e.getMessage());
+                if (e.isUnauthorized()) {
+                    log.info("Re-authenticate and try again");
```
symphony-youri, line 94: as auth expiration is something expected, should we log an error above?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+                    authSession.refresh();
+                }
+                throw e;
```
symphony-youri, line 97: so if auth has expired it means we still throw the exception and go into the retry from resilience4j and wait before retrying the request where we could retry directly

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+                log.error("Error {}: {}", e.getCode(), e.getMessage());
+                if (e.isUnauthorized()) {
+                    log.info("Re-authenticate and try again");
+                    authSession.refresh();
+                }
+                if (e.isClientError()) {
+                    log.info("Recreate a new datafeed and try again");
+                    datafeedId = this.createDatafeedAndSaveToDisk();
+                }
+                throw e;
```
symphony-youri, line 77: same as below 
_so if auth has expired it means we still throw the exception and go into the retry from resilience4j and wait before retrying the request where we could retry directly_

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+    public void start() throws Throwable {
+        this.datafeedId = this.retrieveDatafeedIdFromDisk();
+
+        if (this.datafeedId == null) {
+            this.datafeedId = this.createDatafeedAndSaveToDisk();
+        }
+
+        log.debug("Start reading events from datafeed {}", datafeedId);
+        do {
+            this.started.set(true);
```
symphony-youri, line 46: if I do a start from a thread, and a stop from another thread that is executed before this line then my datafeed won't stop maybe we need some synchronization here

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+        } while (this.started.get());
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void stop() {
+        this.started.set(false);
+        this.datafeedId = null;
```
symphony-youri, line 57: is it needed?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+            this.started.set(true);
+            this.readDatafeed();
+        } while (this.started.get());
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void stop() {
```
symphony-youri, line 55: does stop work if the thread is waiting in datafeedApi.v4DatafeedIdReadGet? should not we try to interrupt it somehow?

symphony-bdk-core/pom.xml
```diff
@@ -97,6 +97,12 @@
             <version>${jackson.version}</version>
         </dependency>
 
+        <dependency>
```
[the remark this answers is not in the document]
symphony-hong: I don't think it will bring a lot of dependencies with it. I used only artifact `resilience4j-retry` and from what I see in [resilience4j-retry](https://mvnrepository.com/artifact/io.github.resilience4j/resilience4j-retry/1.5.0), it doesn't include many dependencies

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -0,0 +1,158 @@
...
+    public void unsubscribe(DatafeedEventListener listener) {
+        listeners.remove(listener);
+    }
+
+    protected void handleV4EventList(List<V4Event> events) {
+        for (V4Event event : events) {
+            if (event == null || event.getType() == null) {
+                continue;
+            }
+            if (event.getInitiator() != null && event.getInitiator().getUser() != null
```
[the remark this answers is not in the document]
symphony-hong: for example, if the bot sends a message, DF service will send an `MESSAGESENT` event to the bot itself, so I guess the bot will not be expected to handle the events which were made by itself (that's what I understand from the legacy sdk, so I applied it to 2.0)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+            fw.flush();
+        } catch (IOException e) {
+            log.error(e.getMessage());
+        }
+    }
+
+    private File getDatafeedIdFile() {
+        String pathToDatafeedIdFile = bdkConfig.getDatafeed().getDatafeedIdFilePath() + DATAFEED_ID_FILE;
+
+        File file = new File(pathToDatafeedIdFile);
```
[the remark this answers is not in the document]
symphony-hong: This function is wrong! I will correct it.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/BdkExponentialFunction.java
symphony-hong: I think the `maxIntervalMillis` is needed to limit the interval in some case that user want to have many many attempts but not exploding the interval by the multiplier.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
symphony-hong: The idea of Darlys is that we have a global retry configuration for all services in the bot, but in each service configuration, we can override this retry config.

symphony-hong: this is global retry config as I answered your comment about the retry configuration in DF config

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
symphony-hong: If I'm not mistaken, the `AtomicBoolean` is thread-safe so the datafeed will stop by a stop from another thread.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
symphony-hong: `DatafeedServiceV1#stop` only breaks the reading loop, the last read will be still finished after it really stops.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
symphony-hong: I'm a bit confused now by the retry strategy. This strategy was suggested by Darlys, and I'm also not sure if we can have two different config in just one Retry instance.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedService.java
symphony-hong: So we could wrap the `Retry#executeCheckedSupplier` with a try/catch and checked the type of `Throwable` which got caught, right?

symphony-bdk-core/pom.xml
```diff
@@ -227,7 +233,8 @@
 
                         <!-- Exclude temporary classes from coverage -->
                         <exclude>com/symphony/bdk/core/SymphonyBdk.class</exclude>
-                        <exclude>com/symphony/bdk/core/service/**/*.class</exclude>
+                        <exclude>com/symphony/bdk/core/service/V4MessageService.class</exclude>
```
symphonydarlys, line 236: why do we exclude this classes?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -51,4 +58,17 @@ public V4MessageService messages(Obo.Handle oboHandle) {
     }
     return new V4MessageService(this.apiClientFactory.getAgentClient(), oboSession);
   }
+
+  public DatafeedService datafeed() {
+    if (DatafeedVersion.of(this.config.getDatafeed().getVersion()) == DatafeedVersion.V2) {
+      return new DatafeedServiceV2(this.apiClientFactory.getAgentClient(), this.apiClientFactory.getPodClient(), this.botSession, this.config);
+    } else {
```
symphonydarlys, line 65: you do not need this else here!!!

symphony-bdk-core/pom.xml
symphony-hong: These classes was created by @symphony-thibault and he left a comment on this class: 

> PLEASE PLEASE don't review this class !!

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -51,4 +58,17 @@ public V4MessageService messages(Obo.Handle oboHandle) {
     }
     return new V4MessageService(this.apiClientFactory.getAgentClient(), oboSession);
   }
+
+  public DatafeedService datafeed() {
```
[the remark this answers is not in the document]
symphony-hong: yep! this PR is still draft, I still have many thing about test and documentation to be added.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorRsaImpl.java
```diff
@@ -51,7 +49,7 @@ public BotAuthenticatorRsaImpl(
    * {@inheritDoc}
    */
   @Override
-  public @Nonnull AuthSession authenticateBot() {
+  public @Nonnull AuthSession authenticateBot() throws AuthUnauthorizedException {
```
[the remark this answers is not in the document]
symphonydarlys: +1

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/legacy/model/LegacySymConfig.java
```diff
@@ -23,26 +23,26 @@
     //
     private String scheme = DEFAULT_SCHEME;
     private String sessionAuthHost;
-    private int sessionAuthPort;
-    private String sessionAuthContextPath;
+    private int sessionAuthPort = 443;
```
symphonydarlys, line 26: why not using a constant for 443?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkConfig.java
symphonydarlys: we need to document that in the readme file, to explain to developer community this approach!!!

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRetryConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private int maxAttempts = DEFAULT_MAX_ATTEMPTS;
```
symphonydarlys, line 15: how this attributes are overridden with the configuration defined by bot developers?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/BotInfoService.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.symphony.bdk.core.api.invoker.ApiClient;
+import com.symphony.bdk.core.api.invoker.ApiException;
+import com.symphony.bdk.core.auth.AuthSession;
+import com.symphony.bdk.gen.api.SessionApi;
+import com.symphony.bdk.gen.api.model.UserV2;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class BotInfoService {
```
symphonydarlys, line 11: documentation for this service please!!!

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedVersion.java
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Get {@link DatafeedVersion} from a String
+     *
+     * @param version version in string
+     * @return DatafeedVersion
+     */
+    public static DatafeedVersion of(String version) {
+        if ("v2".equalsIgnoreCase(version)) {
+            return V2;
+        } else {
```
symphonydarlys, line 15: not needed this else

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -0,0 +1,158 @@
...
+import com.symphony.bdk.gen.api.model.V4Event;
+import io.github.resilience4j.retry.Retry;
+import io.github.resilience4j.retry.RetryConfig;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Slf4j
+abstract class AbstractDatafeedService implements DatafeedService {
```
symphonydarlys, line 23: why do we have this abstract class? doc!!!

symphonydarlys:
unit test are considered?

symphonydarlys:
Close PR by error

symphony-hong:
@symphony-youri My PR is ready for review

symphony-youri (commented):
Running the tests generate a lot of output in the console maybe we need to tweak the log levels to avoid that

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/DatafeedExampleMain.java
```diff
@@ -0,0 +1,42 @@
...
+        final BdkConfig config = BdkConfigLoader.loadFromClasspath("/config.yaml");
+
+        SymphonyBdk bdk = new SymphonyBdk(config);
+        log.info("DatafeedV1: Start");
+        DatafeedService ds = bdk.datafeed();
+
+        ds.subscribe(new DatafeedEventListener() {
+            @Override
+            public void onMessageSent(V4Event event) {
+                if (event.getPayload() != null && event.getPayload().getMessageSent() != null
```
symphony-youri, line 30: as a user should I have to call event.getPayload().getMessageSent().getMessage() or should I be able to just retrieve event.getMessage()?

symphony-youri (commented):
> Running the tests generate a lot of output in the console maybe we need to tweak the log levels to avoid that

Ok thoses are the legacy modules, let's leave them like that

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1Test.java
```diff
@@ -0,0 +1,282 @@
...
+                .thenReturn(events);
+
+        this.datafeedService.unsubscribe(this.listener);
+        this.datafeedService.subscribe(new DatafeedEventListener() {
+            @Override
+            public void onMessageSent(V4Event event) {
+                try {
+                    datafeedService.start();
+                } catch (AuthUnauthorizedException | ApiException e) {
+                    e.printStackTrace();
```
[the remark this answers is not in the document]
symphony-youri: (it might seem less important in tests but if we have the habit of not using it every where it will be easier)
and we also don't want to pollute test outputs with logs and stacktraces

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1Test.java
```diff
@@ -0,0 +1,282 @@
...
+        List<V4Event> events = new ArrayList<>();
+        events.add(null);
+        Field[] fields = DatafeedEventConstant.class.getDeclaredFields();
+        for (Field f : fields) {
+            if (Modifier.isStatic(f.getModifiers()) && Modifier.isPublic(f.getModifiers())) {
+                try {
+                    V4Event event = new V4Event().type((String) f.get(String.class));
+                    events.add(event);
+                } catch (IllegalAccessException e) {
+                    e.printStackTrace();
```
symphony-youri, line 245: to remove too

docs/configuration.md
```diff
@@ -17,6 +17,37 @@ the certificate for authenticating the service account on pod.
 - `app` contains information about the extension app that the bot will use like 
 the appId, the private key or the certificate for authenticating the extension app.
 - `ssl` contains trustStore and trustStore password for SSL communication.
+- `datafeed` contains information of the datafeed service to be used by the bot.
+- `retry` contains information for retry mechanism to be used by the bot.
+
+### Retry Configuration
+The retry mechanism used by the bot will be configured by these following properties:
```
symphony-youri, line 24: How is that going to play with the BOC? (i.e should put in place such a complex configuration if we plan to delegate some of that logic to the BOC)?

symphony-youri (approved):
We can merge as to unblock the rest of the work of DFv2 but it would be interested to get feedback from @symphony-elias and @symphony-thibault 

And it might be worth discussing if exposing v1/v2 to the user is needed (https://perzoinc.atlassian.net/wiki/spaces/DevX/pages/1137084912/6.6.1.1.+BDK2.0+-+Architecture+-+Datafeed+API+Specification?focusedCommentId=1159891949#comment-1159891949)

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/DatafeedExampleMain.java
symphony-hong: In the legacy sdk, as far as I remember, there's several tickets asking for Initiator of an event, therefore, I think user needs more information rather than just a message payload.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/BdkExponentialFunction.java
```diff
@@ -0,0 +1,36 @@
...
+public class BdkExponentialFunction {
+
+    /**
+     * Produce an interval function from given {@link BdkRetryConfig}.
+     *
+     * @param retryConfig given retry configuration.
+     *
+     * @return an {@link IntervalFunction} to be used in Retry mechanism.
+     */
+    public static IntervalFunction ofExponentialBackoff(BdkRetryConfig retryConfig) {
```
symphony-thibault, line 20: Remember that we created a dedicated story (cf. APP-2821) to properly manage retries.

symphony-thibault (approved):
LTGM !

PR #182, opened 2020-08-24
APP-2905: Datafeed Service V2 implementation

symphony-hong:
- Datafeed Service V2 implementation
- Enable Jacoco check on DatafeedServiceV2 with code coverage > 90%
- Documentation already added in V1 PR


symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.io.File;
+
+@Getter
+@Setter
+public class BdkDatafeedConfig {
+
+    private String version = "v1";
```
symphonydarlys, line 12: why this attribute with this value by default?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
```diff
@@ -0,0 +1,185 @@
...
+            this.datafeed = this.retrieveDatafeed();
+            if (this.datafeed == null) {
+                this.datafeed = this.createDatafeed();
+            }
+            log.debug("Start reading datafeed events");
+            do {
+                this.started.set(true);
+                this.readDatafeed();
+            } while (this.started.get());
+        } catch (Throwable e) {
```
symphonydarlys, line 52: it possible to improve this exception catch?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/BdkExponentialFunction.java
```diff
@@ -0,0 +1,36 @@
+package com.symphony.bdk.core.util;
+
+import com.symphony.bdk.core.config.model.BdkRetryConfig;
+import io.github.resilience4j.core.IntervalFunction;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Custom BdkExponentialFunction class to be used in Bdk Retry mechanism.
```
symphonydarlys, line 8: it is not explain the role of this class

symphonydarlys:
All classes must be well documented please. This is a SDK!!!

opoupeney:
Can't agree more with Darlys. Documentation and naming of class/methods has to be crystal clear and consistent.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
symphony-hong: because this class is used to parse config file by` JSONMapper` and this mapper parse version attribute from json or yaml into String so this attribute should be kept as String

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
symphony-hong: because the `Retry#executeCheckedSupplier` throws a `Throwable` then this block of code have to catch `Throwable`.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/BdkExponentialFunction.java
symphony-hong: Okay I will explain it better

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/RealTimeEventListener.java
```diff
@@ -3,9 +3,9 @@
 import com.symphony.bdk.gen.api.model.*;
 
 /**
- * Interface definition for a callback to be invoked when a datafeed event is received.
+ * Interface definition for a callback to be invoked when a real-time event is received.
```
symphony-thibault, line 6: ... "is received from the datafeed."

symphony-thibault: You could also add a link to https://developers.symphony.com/restapi/docs/real-time-events using Javadoc annotation : 
```
/**
 * @see <a href="https://developers.symphony.com/restapi/docs/real-time-events">real-time-events</a>
 */
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -86,117 +86,117 @@ protected void handleV4EventList(List<V4Event> events) throws ApiException {
             if (this.isSelfGeneratedEvent(event)) {
                 continue;
             }
-            for (DatafeedEventListener listener : listeners) {
+            for (RealTimeEventListener listener : listeners) {
                 switch (event.getType()) {
```
symphony-thibault, line 90: We could try to make this part of code a bit more generic by using an `enum` instead of a `switch` (cf. https://github.com/SymphonyOSF/symphony-sdk-java-v2/blob/master/symphony-sdk-api/src/main/java/com/symphony/sdk/api/datafeed/DatafeedEventType.java). We can discuss about that later.

symphony-youri: +1

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/BotInfoService.java
```diff
@@ -8,15 +8,15 @@
...
  * BotInfoService can help the bot retrieving the information about itself by using session API.
  */
 @Slf4j
 public class BotInfoService {
 
-    private SessionApi sessionApi;
     private final AuthSession authSession;
-    private static UserV2 userInfo;
+    private SessionApi sessionApi;
+    private UserV2 userInfo;
```
symphony-youri, line 19: Here the userInfo is retrieved once and then never refreshed, is it ok for the usage we have today?
Is there a risk the user misuses this?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
```diff
@@ -1,36 +1,200 @@
...
+            if (e instanceof ApiException) {
+                throw (ApiException) e;
+            } else if (e instanceof AuthUnauthorizedException) {
+                throw (AuthUnauthorizedException) e;
+            } else {
+                e.printStackTrace();
```
symphony-youri, line 73: to remove

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
symphony-youri: we could have specific catches for each type and then at the end a catch for Throwable
(SonarLint is reporting this as a warning in IDEA, I highly recommend installing this plugin ;) it helps to catch Sonar problems before having it reported in the PR, and it is also reporting more issues given how old the Sonar PR checker is....)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
```diff
@@ -1,36 +1,200 @@
...
+                } else {
+                    log.error("Error {}: {}", e.getCode(), e.getMessage());
+                    if (e.isClientError()) {
+                        try {
+                            log.info("Try to delete the faulty datafeed");
+                            this.deleteDatafeed();
+                            log.info("Recreate a new datafeed and try again");
+                            this.datafeed = this.createDatafeed();
+                        } catch (Throwable throwable) {
+                            e.addSuppressed(throwable);
```
symphony-youri, line 166: woohh first time I see this being used, what does it do?

symphony-hong: If a read datafeed fails and return error 400, it means that the datafeed is stale or faulty, in this case, the bot will have to delete the datafeed and create a new one. And these two steps are also in retry mechanism, it means we have a nested retry here. And if the nested retry failed, it should be stop the parent retry instead of trigger the parent retry to do it again. That's why I add a Suppressed throwable to check if the throwable was thrown by nested retry or not.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/BotInfoService.java
symphony-hong: IMHO, the information of a service account should not be changed during the bot is running. That's why I think it need to be retrieved only once.

symphony-youri: Actually do we need to call the infoservice? don't we have the username available in the bot's configuration and in the event we receive? (to avoid calling the API)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2Test.java
```diff
@@ -0,0 +1,303 @@
...
+    private RealTimeEventListener listener;
+
+    @BeforeEach
+    void setUp() throws BdkConfigException, ApiException {
+        this.bdkConfig = BdkConfigLoader.loadFromClasspath("/config/config.yaml");
+        BdkDatafeedConfig datafeedConfig = this.bdkConfig.getDatafeed();
+        datafeedConfig.setVersion("v2");
+        this.bdkConfig.setDatafeed(datafeedConfig);
+        BdkRetryConfig retryConfig = new BdkRetryConfig();
+        retryConfig.setInitialIntervalMillis(500);
```
symphony-youri, line 45: we can make the tests faster by setting a small value here

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2Test.java
```diff
@@ -0,0 +1,303 @@
...
+        when(datafeedApi.listDatafeed("1234", "1234")).thenReturn(Collections.singletonList(new V5Datafeed().id("test-id")));
+        when(datafeedApi.readDatafeed("test-id", "1234", "1234", ackId)).thenThrow(new ApiException(500, "client-error"));
+
+        assertThrows(ApiException.class, this.datafeedService::start);
+        verify(datafeedApi, times(1)).listDatafeed("1234", "1234");
+        verify(datafeedApi, times(2)).readDatafeed("test-id", "1234", "1234", ackId);
+    }
+
+    @Test
+    void testStartClientErrorDeleteDatafeed() throws ApiException, AuthUnauthorizedException {
```
symphony-youri, line 257: running this test locally shows that the retry is used and at some points it seems to wait for 500ms
but should it be the case for this scenario? we have an invalid DF if, recreate one and wait 500ms before it works again?

symphony-youri (commented):
The retry logic seems a bit difficult to understand and to follow
Would it not be simpler to keep a simple loop as part of start() method, and have a single retry around that (that also checks for this.started before retrying) ?
Also how does it work if we stop the service but have failures and are stuck in the retries somewhere? it means to really stop the service we would have to wait for all retries to finish?

symphony-youri:
Of for instance in the beginning of start, let's say we messed up the authentication

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/BotInfoService.java
symphony-hong: I also thought about it but didn't find another way

symphony-hong:
@symphony-youri  We can wrap a simple start loop by a single retry, but I'm not sure that we expected the retry should be applied to which level (DatafeedService level or API call level). For instance, if we fail to delete a datafeed, do we expect to retry only the delete call or whole start method (retrieve, read...)?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/BotInfoService.java
symphony-thibault: @symphony-hong : in class `AbstractDatafeedService`, could you try to change the method `isSelfGeneratedEvent` by : 

```java
    private boolean isSelfGeneratedEvent(V4Event event) throws ApiException {
        return event.getInitiator().getUser().getUsername().equals(this.bdkConfig.getBot().getUsername());
    }
```

In this case, call to the infoservice is not necessary.

symphony-hong:
And this discussion gives me an idea that we can now remove all the retry mechanism in DatafeedService class. In DatafeedService, we just simply call the api from DatafeedApi, don't care about retry at all, and we will implement the retry mechanism inside the DatafeedApi by templating it using swagger spec.
What do you think? @symphony-youri @symphony-thibault

symphony-youri:
Maybe it would be ok just restart the whole process, it could help to make the code easier to read
But no blocker for me on this, this is internal so we can test it further and improve it later

symphony-youri:
what!!! di

symphony-thibault:
@symphony-youri we have a dedicated story for next Sprint to improve the retry mechanism. Actually, implementing retries shouldn't have been part of this PR to make review simpler and shorter. We decided with @symphony-hong to let it as it is and fix it on next Sprint

symphony-youri:
> And this discussion gives me an idea that we can now remove all the retry mechanism in DatafeedService class. In DatafeedService, we just simply call the api from DatafeedApi, don't care about retry at all, and we will implement the retry mechanism inside the DatafeedApi by templating it using swagger spec.
> What do you think? @symphony-youri @symphony-thibault

Isn't it going to the caller of the API that can decide of the retry mechanism to use?

symphony-youri (approved):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
```diff
@@ -1,36 +1,196 @@
...
+            throw new IllegalStateException("The datafeed service is already started");
+        }
+        try {
+            this.datafeed = this.retrieveDatafeed();
+            if (this.datafeed == null) {
+                this.datafeed = this.createDatafeed();
+            }
+            log.debug("Start reading datafeed events");
+            do {
+                this.started.set(true);
```
symphony-elias, line 64: Do we want to set `started` to true inside the loop? IMHO it would make more sense to set it once just before the loop.

symphony-hong: Oops! I made mistake.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -57,14 +74,10 @@ public void start() throws AuthUnauthorizedException, ApiException {
...
-                throw (ApiException) e;
-            } else if (e instanceof AuthUnauthorizedException) {
-                throw (AuthUnauthorizedException) e;
-            } else {
-                log.error("Unknown error", e);
-            }
+        } catch (AuthUnauthorizedException | ApiException exception) {
+            throw exception;
+        } catch (Throwable throwable) {
+            log.error("Unknown error", throwable);
```
symphony-elias, line 80: Do we know what kind of errors it could be ? Do we have a reason why we want to silently ignore these errors ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
```diff
@@ -1,36 +1,196 @@
...
+                this.readDatafeed();
+            } while (this.started.get());
+        } catch (AuthUnauthorizedException | ApiException exception) {
+            throw exception;
+        } catch (Throwable throwable) {
+            log.error("Unknown error", throwable);
```
symphony-elias, line 70: Same thought as put in  symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
symphony-thibault: +1

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV2.java
```diff
@@ -1,36 +1,196 @@
...
+            try {
+                return this.datafeedApi.createDatafeed(authSession.getSessionToken(), authSession.getKeyManagerToken());
+            } catch (ApiException e) {
+                if (e.isUnauthorized()) {
+                    log.info("Re-authenticate and try again");
+                    authSession.refresh();
+                } else {
+                    log.error("Error {}: {}", e.getCode(), e.getMessage());
+                }
+                throw e;
```
symphony-elias, line 99: If `e.isUnauthorized()` it means we will execute `authSession.refresh()` and throw the error. Is this intended ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -59,155 +53,154 @@ public AbstractDatafeedService(ApiClient agentClient, ApiClient podClient, AuthS
...
                     case DatafeedEventConstant.MESSAGESENT:
                         listener.onMessageSent(event);
                         if (event.getPayload() != null) {
-                            listener.onMessageSent(event.getPayload().getMessageSent());
+                            listener.onMessageSent(event.getInitiator(), event.getPayload().getMessageSent());
                         }
                         break;
```
symphony-thibault, line 89: This `if` condition is not necessary. Our API contract ensure that this field will always be part of the response payload so I'd modify this `case` (as well as the following ones) to : 
```java
case: 
    listener.onMessageSent(event);
    listener.onMessageSent(event.getInitiator(), event.getPayload().getMessageSent());
    break;
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
symphony-hong: AuthUnauthorizedException and ApiException exception are the only two exception can happen in our case. In spite of it, the `Retry#executeCheckedSupplier` still throws a Throwable because it didn't care about what we did inside the retried block. That's why we have this catch here, although in fact, it is never caught.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
symphony-hong: Generated class marked method `getPayload` with annotation `@Nullable`, so it created a warning about NullPointerException and in Intellij, it looks quite annoying. 😞

symphony-thibault: Moving to an _enum-based_ pattern (cf. [here](https://github.com/SymphonyOSF/symphony-sdk-java-v2/blob/master/symphony-sdk-api/src/main/java/com/symphony/sdk/api/datafeed/DatafeedEventType.java#L15)) would fix this issue. 

I start having some doubts regarding the fact that we should return the full `V4Event`. This object contains all the different payloads (`onMessageSent`, `onMessageSuppressed`, etc.) and this might create some confusion for some developers.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/RealTimeEventType.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.function.Function;
+
+/**
+ * Enumeration of possible types of Real Time Events that can be retrieved from the DataFeed.
+ * More information : https://rest-api.symphony.com/reference#real-time-events-v4
+ */
+enum RealTimeEventType {
+
+    MESSAGESENT((listener, event, innerEvent) -> {
+        listener.onMessageSent(event);
```
symphony-thibault, line 18: As I mentioned earlier, I don't think that we should consider providing the whole `V4Event` to listener callbacks, this might introduce more confusion to end-developers as well as making our API more difficult to understand. 

The goal of the BDK is really to hide the complexity of our REST API, and this specifically one point that we need to take care of. 

WDYT?

symphony-thibault: Also (and assuming that we would keep this strategy), the `innerEvent` is already part of the `event` so there is no really need to have it part of the lambda function.

symphony-hong: So I think it's better if we keep just one callbacks take the initiator and the dedicated event as inputs. I will remove the other.

symphony-elias (approved):
LGTM

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/BotAuthenticatorRsaImpl.java
```diff
@@ -76,7 +78,7 @@ protected String doRetrieveToken(ApiClient client) throws AuthUnauthorizedExcept
       log.debug("{} successfully retrieved.", token.getName());
       return token.getToken();
     } catch (ApiException ex) {
-      if (ex.getCode() == HttpURLConnection.HTTP_UNAUTHORIZED) {
+      if (ex.getCode() == 401) {
```
symphony-thibault, line 81: I should be `HttpURLConnection.HTTP_UNAUTHORIZED` instead of `401`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -59,155 +53,47 @@ public AbstractDatafeedService(ApiClient agentClient, ApiClient podClient, AuthS
...
-                            listener.onSymphonyElementsAction(event.getPayload().getSymphonyElementsAction());
-                        }
-                        break;
+            try {
+                RealTimeEventType eventType = RealTimeEventType.valueOf(event.getType());
+                for (RealTimeEventListener listener : listeners) {
+                    eventType.dispatch(listener, event);
                 }
+            } catch (IllegalArgumentException e) {
+                log.warn("Receive events with unknown type");
```
symphony-thibault, line 88: That would be nice to log the type of the unknown event as well

symphony-thibault: `log.warn("Receive events with unknown type : {}", event.getType());`

symphony-thibault (approved):
LGTM 🙌 This PR was a big journey :)

PR #183, opened 2020-08-25
Add a Connections Listener to BDK

anthony-symphony:
Add a Connections Listener to BDK
Fix some Windows based tests


symphony-thibault (approved):
LGTM

symphony-bdk-legacy/symphony-bdk-bot-sdk-java/src/main/java/com/symphony/bdk/bot/sdk/event/model/ConnectionAcceptedEvent.java
```diff
@@ -0,0 +1,24 @@
+/*
```
symphony-youri, line 1: Please change your default template as to not have those comments :)

symphony-youri:
Whats the use case for this listener? (just to get a bit more context and understand why it is useful, could be in the commit message to retrieve it easily by looking at the history)

anthony-symphony:
Added reasoning and removed the comments. Reasoning is to allow bots that need to communicate externally to have the ability to listen to incoming connection requests and accept them.

PR #185, opened 2020-08-26
CES-1719: Added the images field to Attachment

anandpadiya:
## Description
- [Jira Ticket](https://perzoinc.atlassian.net/browse/CES-1719)

## Implementation
- We want to make use of the thumbnail of the image attachment.
- We can see that the agent is already sending that inside the attachment object in the `images` field.
- This field was missing in the `Attachment` class. So, we have added it.


symphony-thibault (approved):
LGTM

PR #187, opened 2020-08-27
APP-2970: Fixed infinite call loop in SymConfig

symphony-elias:


symphony-bdk-legacy/symphony-api-client-java/src/main/java/configuration/SymConfig.java
```diff
@@ -101,7 +101,7 @@
     public String getAgentUrl() {
         String port = (this.getAgentPort() == 443) ? "" : ":" + this.getAgentPort();
         String contextPath = formatContextPath(this.getAgentContextPath());
-        return CommonConstants.HTTPS_PREFIX + this.getAgentHost() + port + contextPath;
+        return CommonConstants.HTTPS_PREFIX + this.agentHost + port + contextPath;
```
symphony-youri, line 104: isn't changing the logic? i.e the agent url won't be "load balanced" anymore?

symphony-youri (commented):
It would be nice to add a test reproducing the bug and showing it has been fixed

symphony-bdk-legacy/symphony-api-client-java/src/main/java/configuration/SymConfig.java
symphony-elias: Truth is, I don't know the exact logic here and how to remove the stack overflow error (see my comment on issue APP-2970). So, please advise.

symphony-thibault:
@symphony-elias We (or I?) recently refactored the `SymConfig` class to use Lombok for getters and setters. Could it be related?

symphony-elias:
> @symphony-elias We (or I?) recently refactored the `SymConfig` class to use Lombok for getters and setters. Could it be related?

I guess the related change is this one : https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/commit/b110ceeaa72af78c7d4487d52d0f0e6f57ebbdf5#diff-fc189cefbad045bb241cea9622a1f008

symphony-bdk-legacy/symphony-api-client-java/src/main/java/configuration/SymConfig.java
symphony-youri: @symphony-elias with this test I was able to reproduce the stack overflow error:

```
@Test
  public void getAgentHost_stackOverflow() {
    final SymLoadBalancedConfig config = new SymLoadBalancedConfig();
    final LoadBalancing loadBalancing = new LoadBalancing();
    loadBalancing.setMethod(LoadBalancingMethod.external);
    config.setLoadBalancing(loadBalancing);

    assertNotNull(config.getAgentHost());
  }
```

Not sure this is the exact same configuration as reported in the ticket but this only should not produce a stack overflow

symphony-youri: And with your fix I get 

java.lang.IndexOutOfBoundsException: Index: 0

	at java.util.Collections$EmptyList.get(Collections.java:4456)
	at configuration.SymLoadBalancedConfig.rotateAgent(SymLoadBalancedConfig.java:86)
	at configuration.SymLoadBalancedConfig.getAgentHost(SymLoadBalancedConfig.java:61)

Will there be only one and only one agentServers defined if the LB method is external?
maybe in that case we can improve the error message if it is not defined

symphony-elias: @symphony-youri Yes we expect to have one and only one item (clarified with Philippe Vialatte). I will add unit tests and add an explicit error message.

symphony-youri (approved):

symphony-hong (approved):
LGTM!

PR #190, opened 2020-08-27
Fixed an issue where bot will not respond to any commands 

anthony-symphony:
if there is a null commandMatcher bots will not respond to any commands even if commandMatcher is not null


symphony-elias (approved):
LGTM

PR #191, opened 2020-08-28
APP-2998 Make OBO configuration optional

symphony-elias:
Don't know how to test SymphonyBdk as it has many dependencies.

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
@@ -19,6 +19,8 @@ import com.symphony.bdk.gen.api.MessagesApi;
 import lombok.extern.slf4j.Slf4j;
 import org.apiguardian.api.API;
 
+import java.util.Optional;
+
 /**
  * BDK entry point.
  */
@@ -55,7 +57,7 @@ public class SymphonyBdk {
     return new MessageService(new MessagesApi(this.apiClientFactory.getAgentClient()), this.botSession);
   }
 
-  public MessageService messages(Obo.Handle oboHandle) throws AuthUnauthorizedException, AuthInitializationException {
+  public MessageService messages(Obo.Handle oboHandle) throws AuthUnauthorizedException {
     AuthSession oboSession;
     if (oboHandle.hasUsername()) {
       oboSession = this.getOboAuthenticator().authenticateByUsername(oboHandle.getUsername());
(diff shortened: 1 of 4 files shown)


symphony-youri:
> Don't know how to test SymphonyBdk as it has many dependencies.

Is the problematic dependency the ApiClientFactory? In that case I would create a specific constructor (not public) that takes it as a parameter to inject a mock

symphony-elias:
> > Don't know how to test SymphonyBdk as it has many dependencies.
> 
> Is the problematic dependency the ApiClientFactory? In that case I would create a specific constructor (not public) that takes it as a parameter to inject a mock

Indeed but it means we will have to either duplicate code in the constructor or remove the `final` keywords on the fields.

symphony-youri:
Like that it should be ok (we won't be able to get 100% line coverage though :))
```
public SymphonyBdk(BdkConfig config) throws AuthInitializationException, AuthUnauthorizedException {
    this(config, new ApiClientFactory(config));
  }

  SymphonyBdk(BdkConfig config, ApiClientFactory apiClientFactory)
      throws AuthInitializationException, AuthUnauthorizedException {

    this.apiClientFactory = apiClientFactory;

    final AuthenticatorFactory authenticatorFactory = new AuthenticatorFactory(config, this.apiClientFactory);

    this.botSession = authenticatorFactory.getBotAuthenticator().authenticateBot();
    this.oboAuthenticator = authenticatorFactory.getOboAuthenticator();

    // setup the datafeed
    final DatafeedApi datafeedApi = new DatafeedApi(this.apiClientFactory.getAgentClient());
    if (DatafeedVersion.of(config.getDatafeed().getVersion()) == DatafeedVersion.V2) {
      this.datafeedService = new DatafeedServiceV2(datafeedApi, this.botSession, config);
    } else {
      this.datafeedService = new DatafeedServiceV1(datafeedApi, this.botSession, config);
    }
  }
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -74,4 +74,11 @@ public MessageService messages(Obo.Handle oboHandle) throws AuthUnauthorizedExce
   public DatafeedService datafeed() {
     return this.datafeedService;
   }
+
+  protected OboAuthenticator getOboAuthenticator() throws AuthInitializationException {
+    if (this.oboAuthenticator == null) {
+      throw new AuthInitializationException("OBO is not configured", null);
+    }
+    return this.oboAuthenticator;
+  }
```
symphony-thibault, line 83: I'd write something more Java8 :) 
```java
  protected OboAuthenticator getOboAuthenticator() {
    return Optional.ofNullable(this.oboAuthenticator)
        .orElseThrow(() -> new IllegalStateException("OBO is not configured."));
  }
```

As well, the `AuthInitializationException` is not really designed to handle this kind of error. Either we create a dedicated exception, or we keep the `IllegalStateException`. Anyway, that's the kind of error that a developer should see right when starting the application

symphony-youri: At the cost of allocating memory for the optional! :p

symphony-thibault: Aren't you ex-C++ dev?!

symphony-thibault (approved):
LGTM 👍

PR #192, opened 2020-08-28
APP-2996: Datafeed ID repository for Datafeed Service V1

symphony-hong:
Create an interface DatafeedRepository for handling storing
a datafeed id which is created by using DatafeedServiceV1.

Create OnDiskDatafeedRepository implementing the DatafeedRepository
for storing datafeed id on disk.

In DatafeedServiceV1, users can use the OnDiskDatafeedRepository
implementation by default or we can inject their own repository, for
example: MongoDB, SQL storage...

Update unittest

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedRepository.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedRepository.java
@@ -1,0 +1,22 @@
+package com.symphony.bdk.core.service.datafeed;
+
+/**
+ * A repository interface for storing a datafeed id.
+ * By using DatafeedServiceV1, the created datafeed id should be persisted manually on the BDK side.
+ */
+public interface DatafeedRepository {
+
+  /**
+   * Persists the created datafeed id into the storage.
+   *
+   * @param datafeedId the datafeed id to be persisted.
+   */
+  void write(String datafeedId);
+
+  /**
+   * Read the persisted datafeed id from the storage.
+   *
+   * @return The persisted datafeed id.
+   */
+  String read();
+}
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java
@@ -1,7 +1,7 @@
 package com.symphony.bdk.core.service.datafeed.impl;
 
 import com.symphony.bdk.core.config.model.BdkConfig;
-import com.symphony.bdk.core.service.datafeed.DatafeedRepository;
+import com.symphony.bdk.core.service.datafeed.DatafeedIdRepository;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
@@ -13,18 +13,19 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.Optional;
 
 /**
- * The implementation of Repository interface for persisting a datafeed id on disk.
+ * The implementation of {@link DatafeedIdRepository} interface for persisting a datafeed id on disk.
  */
 @Slf4j
-public class OnDiskDatafeedRepository implements DatafeedRepository {
+public class OnDiskDatafeedIdRepository implements DatafeedIdRepository {
 
   private static final String DATAFEED_ID_FILE = "datafeed.id";
 
   private final BdkConfig config;
 
-  public OnDiskDatafeedRepository(BdkConfig config) {
+  public OnDiskDatafeedIdRepository(BdkConfig config) {
     this.config = config;
   }
 
@@ -46,23 +47,24 @@ public class OnDiskDatafeedRepository implements DatafeedRepository {
    * {@inheritDoc}
    */
   @Override
-  public String read() {
+  public Optional<String> read() {
     log.debug("Reading datafeed id from file: {}", this.getDatafeedIdFile().toString());
-    String datafeedId = null;
+    String datafeedId;
     try {
       File file = this.getDatafeedIdFile();
       Path datafeedIdPath = Paths.get(file.getPath());
       List<String> lines = Files.readAllLines(datafeedIdPath);
       if (lines.isEmpty() || !lines.get(0).contains("@")) {
-        return null;
+        return Optional.empty();
       }
       String[] persistedDatafeed = lines.get(0).split("@");
       datafeedId = persistedDatafeed[0];
       log.info("Retrieved datafeed id from datafeed repository: {}", datafeedId);
+      return Optional.of(datafeedId);
     } catch (IOException e) {
       log.debug("No persisted datafeed id could be retrieved from disk");
+      return Optional.empty();
     }
-    return datafeedId;
   }
 
   private File getDatafeedIdFile() {
--- a/symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
+++ b/symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
@@ -28,23 +28,24 @@ import java.util.Optional;
  */
 public class OnDiskDatafeedIdRepositoryTest {
 
-  private BdkConfig bdkConfig;
   private Path datafeedFile;
+  private DatafeedIdRepository datafeedIdRepository;
 
   @BeforeEach
   void setUp(@TempDir Path tempDir) throws BdkConfigException {
-    this.bdkConfig = BdkConfigLoader.loadFromClasspath("/config/config.yaml");
-    BdkDatafeedConfig datafeedConfig = this.bdkConfig.getDatafeed();
+    BdkConfig bdkConfig = BdkConfigLoader.loadFromClasspath("/config/config.yaml");
+    BdkDatafeedConfig datafeedConfig = bdkConfig.getDatafeed();
     datafeedConfig.setIdFilePath(tempDir.toString());
-    this.bdkConfig.setDatafeed(datafeedConfig);
+    bdkConfig.setDatafeed(datafeedConfig);
     this.datafeedFile = tempDir.resolve("datafeed.id");
+    this.datafeedIdRepository = new OnDiskDatafeedIdRepository(bdkConfig);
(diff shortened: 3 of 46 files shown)


symphony-elias (approved):
LGTM

symphony-youri (commented):
Do we have users that need that? To me it looks like the datafeed id is something quite volatile and that storing it on disk is enough?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedRepository.java
```diff
@@ -0,0 +1,22 @@
...
+   * @param datafeedId the datafeed id to be persisted.
+   */
+  void write(String datafeedId);
+
+  /**
+   * Read the persisted datafeed id from the storage.
+   *
+   * @return The persisted datafeed id.
+   */
+  String read();
```
symphony-thibault, line 21: Could we make this `read` method return an `Optional`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedRepository.java
```diff
@@ -0,0 +1,22 @@
+package com.symphony.bdk.core.service.datafeed;
+
+/**
+ * A repository interface for storing a datafeed id.
+ * By using DatafeedServiceV1, the created datafeed id should be persisted manually on the BDK side.
+ */
+public interface DatafeedRepository {
```
symphony-thibault, line 7: `DatafeedIdRepository`

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedRepositoryTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class OnDiskDatafeedRepositoryTest {
```
symphony-thibault, line 24: Add javadoc comment: 
```
/**
 * Test class for the {@link OnDiskDatafeedRepository}.
 */
```

symphony-thibault:
> @symphony-youri Do we have users that need that? 

Apart from me: no :) But I already know that it would be interesting for the Integration team that already works with MongoDB on on-prem side (future `MongoDatafeedIdRepository`?). 

> To me it looks like the datafeed id is something quite volatile and that storing it on disk is enough?

The datafeed id is a critical information. Loosing it can lead to a loss of messages, and potential business impact for our customers. As well, I'm not sure that the disk strategy works well in a Docker environment, to be confirmed.

Last argument: separation of concerns, simplification of the `DatafeedServiceV1` implementation. Code is cleaner an easier to maintain (pure personal POV though).

opoupeney:
The disk strategy works for most of our customers although it requires to mount a persistent volume in your container (not ideal), that's why the Integration team is looking for a DB alternative (which has its own constraints with the highly secure authentication some customers have)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+
+/**
+ * Test class for the {@link OnDiskDatafeedIdRepository}.
```
symphony-youri, line 27: The comment here is not needed IMHO

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
```diff
@@ -0,0 +1,84 @@
...
+  void setUp(@TempDir Path tempDir) throws BdkConfigException {
+    this.bdkConfig = BdkConfigLoader.loadFromClasspath("/config/config.yaml");
+    BdkDatafeedConfig datafeedConfig = this.bdkConfig.getDatafeed();
+    datafeedConfig.setIdFilePath(tempDir.toString());
+    this.bdkConfig.setDatafeed(datafeedConfig);
+    this.datafeedFile = tempDir.resolve("datafeed.id");
+  }
+
+  @Test
+  void writeDatafeedTest() throws IOException {
```
symphony-youri, line 44: tests are easier to read when steps (Given/When/Then) are separated with line breaks IMHO

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
```diff
@@ -0,0 +1,84 @@
...
+  void readEmptyDatafeedFileTest() throws IOException {
+    FileUtils.writeStringToFile(datafeedFile.toFile(), "", StandardCharsets.UTF_8);
+    DatafeedIdRepository repository = new OnDiskDatafeedIdRepository(bdkConfig);
+    Optional<String> datafeedId = repository.read();
+    assertFalse(datafeedId.isPresent());
+  }
+
+  @Test
+  void readNotExistDatafeedFileTest() {
+    DatafeedIdRepository repository = new OnDiskDatafeedIdRepository(bdkConfig);
```
symphony-youri, line 79: I would perhaps move the init of the repo to the setup as we always do it the same way

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java
```diff
@@ -0,0 +1,77 @@
...
+      List<String> lines = Files.readAllLines(datafeedIdPath);
+      if (lines.isEmpty() || !lines.get(0).contains("@")) {
+        return Optional.empty();
+      }
+      String[] persistedDatafeed = lines.get(0).split("@");
+      datafeedId = persistedDatafeed[0];
+      log.info("Retrieved datafeed id from datafeed repository: {}", datafeedId);
+      return Optional.of(datafeedId);
+    } catch (IOException e) {
+      log.debug("No persisted datafeed id could be retrieved from disk");
```
symphony-youri, line 65: in that case is it ok to hide the exception or should we propagate it?
if it were a database as the repository would it change?

symphony-youri:
> 
> Last argument: separation of concerns, simplification of the `DatafeedServiceV1` implementation. Code is cleaner an easier to maintain (pure personal POV though).

We could do that having a public interface and adding an extension point in the SDK


Isn't going to be quickly obsolete with DF2? Or is going to take a long time to have it deployed everywhere?

symphony-youri:
> 
> The datafeed id is a critical information. Loosing it can lead to a loss of messages, and potential business impact for our customers. As well, I'm not sure that the disk strategy works well in a Docker environment, to be confirmed.


Does it mean we can only have one DF id for a bot? for me it was more like a session token given that we sometime recreate it
Of for instance what happens if we run two instances of the bot with different ids (or the same id?)

opoupeney:
From an agent point of view, a bot can have up to 5 dafateed opened (to be verified though)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
symphony-thibault: I like it (personally), quite convenient when navigating through the code. You can easily jump to the tested class by CTRL+Click on the link. Could be optional, I agree.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java
symphony-hong: I'm not sure if it's needed to be logged or not, so I just log it as debug log. But at least, I think the bot developers should be notified if no datafeed id is retrieved.

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
symphony-youri: command + shit + t :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java
symphony-youri: For instance let's say it is a database and the database is not available, could we end creating datafeed indefinitely?
Or that we don't have write permissions on the file?

symphony-youri (approved):

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepositoryTest.java
symphony-thibault: Screenshot 2020-08-28 at 18 56 54

Did not know this one !

symphony-thibault (commented):
LGTM 👍

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java
symphony-hong: Oh so, you're talking about the one in write method. Maybe we should propagate the exception? But if we throws IOException, then we might have problem when we implement other kinds of repository. Maybe defining a dedicated exception might work?

symphony-youri: I'd say both read/write could raise exceptions, IOException would be fine by me

PR #193, opened 2020-08-31
First implementation of templates

symphony-elias:


symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/core/template/api/Template.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+public interface Template {
+
+  /**
+   *
+   * @param parameters the map which contains the parameter names as map key and the parameter values as map values
+   * @return the generated string where each parameter is replaced by its value provided in the map
+   * @throws TemplateException in case of issues during the string generation, e.g. missing parameter
+   */
+  String process(Map<String, String> parameters) throws TemplateException;
```
symphony-thibault: With a `Map<String, String>` we can only handle 1 fields level in templates. What happen if a template engine accepts complex objects (`${user.address.street}` for example)?

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/Template.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+public interface Template {
+
+  /**
+   *
+   * @param parameters the map which contains the parameter names as map key and the parameter values as map values
+   * @return the generated string where each parameter is replaced by its value provided in the map
+   * @throws TemplateException in case of issues during the string generation, e.g. missing parameter
+   */
+  String process(Map<String, String> parameters) throws TemplateException;
```
symphony-thibault, line 17: With a `Map<String, String>` as parameter we can only manage 1 level of attributes when most of the template engines support tree values as data-model (cf. https://freemarker.apache.org/docs/dgui_quickstart_datamodel.html).

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/core/template/api/Template.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Map;
+
+/**
+ * Interface to represent a template.
+ * A template takes parameters in input and outputs a string.
+ */
+public interface Template {
+
+  /**
+   *
```
symphony-thibault: Overall description missing

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/core/template/api/Template.java
```diff
@@ -0,0 +1,18 @@
+package com.symphony.bdk.core.template.api;
+
+import java.util.Map;
+
+/**
+ * Interface to represent a template.
+ * A template takes parameters in input and outputs a string.
+ */
+public interface Template {
```
symphony-thibault: `@API` Guardian annotation are missing

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/Template.java
symphony-thibault: Freemarker can process an object for instance (https://freemarker.apache.org/docs/api/freemarker/template/Template.html#process-java.lang.Object-java.io.Writer-), that would be nice to study different template engines to see what they offer but I think this is where we should go

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/core/template/api/Template.java
symphony-elias: 👍

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/Template.java
symphony-elias: 👍

symphony-elias: FreeMarker accepts any kind of objects, like Map<String, Object> or java beans: https://freemarker.apache.org/docs/pgui_quickstart_createdatamodel.html
Velocity accepts a Context object which is a Map like object, see : http://velocity.apache.org/engine/devel/developer-guide.html#the-context
I would suggest to accept a Object in the interface and each implementation will have to check the actual type.

symphony-thibault (commented):
+ create a demo in `sumphony-bdk-examples` module

symphony-bdk-template/symphony-bdk-template-freemarker/src/main/java/com/symphony/bdk/template/freemarker/UrlTemplateLoader.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+/**
+ * Internal class used when instantiating a {@link FreeMarkerTemplate} with {@link FreeMarkerEngine#newTemplateFromUrl(String)}
+ */
+public class UrlTemplateLoader extends URLTemplateLoader {
```
symphony-hong, line 13: I think this class should be package-protected. I don't think developers will touch this class.

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/TemplateEngine.java
```diff
@@ -0,0 +1,73 @@
...
+            .collect(Collectors.toList());
+
+    if (templateEngines.isEmpty()) {
+      throw new IllegalStateException("No TemplateEngine implementation found in classpath.");
+    } else if (templateEngines.size() > 1) {
+      LoggerFactory.getLogger(TemplateEngine.class)
+          .warn("More than 1 TemplateEngine implementation found in classpath, will use : {}",
+          templateEngines.stream().findFirst().get());
+    }
+    LoggerFactory.getLogger(TemplateEngine.class).info("TemplateEngine implementation found");
```
symphony-elias, line 70: @symphony-thibault I cannot use the Slf4j annotation as it is only valid on classes and enums. This leads to the following error :  Failed to load class "org.slf4j.impl.StaticLoggerBinder".
What should I do ?

symphony-bdk-examples/bdk-template-examples/src/main/java/com/symphony/template/examples/FreeMarkerExample.java
```diff
@@ -0,0 +1,33 @@
...
+ * Sample class showing usage of FreeMarker templates.
+ */
+public class FreeMarkerExample {
+  public static void main(String[] args) throws TemplateException {
+    System.out.println(getStringFromBuiltInTemplate());
+    System.out.println(getStringFromCustomTemplate());
+  }
+
+  private static String getStringFromBuiltInTemplate() throws TemplateException {
+    Template template = new FreeMarkerEngine().newBuiltInTemplate("simpleMML");
```
symphony-hong, line 19: Do you think we need a enum listing all built-in templates we have? IMHO, it might be more convenient if we write something like: ```new FreeMarkerEngine().newBuiltInTemplate(BuiltInTemplate.SIMPLE_MML);```
WDYT?

symphony-thibault (approved):
LGTM 👍

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/TemplateEngine.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  Template newBuiltInTemplate(String template) throws TemplateException;
+
+  /**
+   * Create a {@link Template} instance from a file on the file system
+   * @param templatePath path to a template file on the file system
+   * @return a new {@link Template} instantiated from the provided file
+   * @throws TemplateException when template cannot be loaded, e.g. file not accessible
+   */
+  Template newTemplateFromFile(String templatePath) throws TemplateException;
```
symphony-youri, line 39: perhaps we could support URI? and then find out it is a file or the classpath based on it?

file:///root/blah.text classpath://com/symphony/bdk/blah.text

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/TemplateEngine.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  Template newTemplateFromClasspath(String templatePath) throws TemplateException;
+
+  /**
+   * Creates a template from a URL to a template file
+   * @param url the url where to fetch the template file
+   * @return a new {@link Template} instantiated from the provided url, should be a valid {@link java.net.URL} string.
+   * @throws TemplateException when template cannot be loaded, e.g. url not accessible
+   */
+  Template newTemplateFromUrl(String url) throws TemplateException;
```
symphony-youri, line 55: ah we have URL here, would not URI cover everything?

PR #195, opened 2020-09-01
APP-2983: Implementation of user service in BDK 2.0

symphony-hong:
Create UserService class to wrap all the user api endpoints provided
by the pod to retrieve user, list of users, perform action to get and
update role, delegate, entitlements, avatar, disclaimer, status of user.

Create an entry to user service in SymphonyBdk by calling method SymphonyBdk#user

Create an example for user service

Update javadoc

Unittest added with code coverage >90%


symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -29,23 +33,28 @@
...
 
   private final ApiClientFactory apiClientFactory;
+  private final ApiClient podClient;
+  private final ApiClient agentClient;
 
   private final AuthSession botSession;
   private final OboAuthenticator oboAuthenticator;
 
   private final DatafeedService datafeedService;
+  private UserService userService;
```
symphony-thibault, line 43: Could be final and directly initialized within the main constructor

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserDetailMapper.java
```diff
@@ -0,0 +1,60 @@
+package com.symphony.bdk.core.service.user;
+
+import com.symphony.bdk.gen.api.model.UserAttributes;
+import com.symphony.bdk.gen.api.model.UserDetail;
+import com.symphony.bdk.gen.api.model.V2UserAttributes;
+import com.symphony.bdk.gen.api.model.V2UserDetail;
+
+class UserDetailMapper {
```
symphony-thibault, line 8: Could we use https://mapstruct.org/ ?

symphony-hong: Updated but I need your review on pom file because the build on CircleCI failed on maven-compiler!

symphony-thibault: Is failing locally as well?

symphony-thibault:
@symphony-hong How do the services will handle OBO authentication? At the moment, it is only possible to call those services using the regular bot authentication (i.e. bot's `AuthSession`). I think that we should also provide prototypes that accept an `AuthSession` as parameter. WDYT?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserDetailMapper.java
symphony-hong: it failed on local as well

symphony-hong:
> How do the services will handle OBO authentication?

I thought, for handling the OBO authentication, we just need pass the obo's `AuthSession` to the constructor like what is being implemented in MessagaeService, isn't it?

symphony-thibault:
> I thought, for handling the OBO authentication, we just need pass the obo's AuthSession to the constructor like what is being implemented in MessageService, isn't it?

`MessageService` was more a POC than a definitive implementation. The problem with this design is that we need to create a new instance each time we want to call in OBO mode. Which can be very costly in terms of memory if an application has to manage a lot of OBO sessions.

symphony-hong:
I have several ideas for it:
- Take the authentication out of the SymphonyBdk, accept `AuthSession` as parameter
- Create service for both normal and obo mode, store in hashmap (Flyweight Pattern)

symphony-thibault:
@symphony-hong could you make a small demo for both of each?

symphony-hong:
> could you make a small demo for both of each?
Okay I will do it when we're working with the ticket about Fluent API

symphony-hong:
@symphony-thibault or we can do like what I did in last update. Is this your point at the beginning?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,405 @@
...
+ *   <li>Get, update status of an user</li>
+ * </ul></p>
+ *
+ */
+@Slf4j
+public class UserService {
+
+  private final UserApi userApi;
+  private final UsersApi usersApi;
+  private AuthSession authSession;
```
symphony-thibault, line 51: In a multi-threaded environment (Spring for instance): this will cause big issues for sure :) Also, we should also keep in mind that OBO is not supported for all API endpoints (cf. https://developers.symphony.com/restapi/reference#obo-enabled-endpoints). 

We should enforce developers to use OBO authentication on _OBO-Enabled endpoints_ only.

symphony-bdk-core/src/test/java/com/symphony/bdk/core/test/BdkMockServer.java
```diff
@@ -51,6 +52,10 @@ public void onGet(String path, Consumer<HttpResponse> resModifier) {
     this.onRequest("GET", path, resModifier);
   }
 
+  public void onDelete(String path, Consumer<HttpResponse> resModifier) {
+    this.onRequest("DELETE", path, resModifier);
+  }
+
   public void onRequest(String method, String path, Consumer<HttpResponse> resModifier) {
```
symphony-elias, line 59: Please factorize that with your `onRequestFailed` function

symphony-elias (approved):
LGTM

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserDetailMapper.java
```diff
@@ -0,0 +1,15 @@
+package com.symphony.bdk.core.service.user;
+
+import com.symphony.bdk.gen.api.model.UserDetail;
+import com.symphony.bdk.gen.api.model.V2UserDetail;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper
+public interface UserDetailMapper {
```
symphony-thibault, line 10: This class could be moved to a sub-package called `mapper`

pom.xml
```diff
@@ -17,6 +17,8 @@
 
         <!-- Plugins -->
         <jacoco-maven-plugin.version>0.8.5</jacoco-maven-plugin.version>
+        <mapstruct.version>1.3.1.Final</mapstruct.version>
```
symphony-thibault, line 20: `<mapstruct-plugin.version>` and `<lombok-plugin.version>`. Just to easily differentiate dependencies and plugins

symphony-bdk-bom/pom.xml
```diff
@@ -22,6 +22,7 @@
         <commons-io.version>2.6</commons-io.version>
         <commons-lang3.version>3.9</commons-lang3.version>
         <lombok.version>1.18.12</lombok.version>
+        <mapstruct.version>1.3.1.Final</mapstruct.version>
```
symphony-thibault, line 25: **Alphabetical order** :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
+package com.symphony.bdk.core.service.user;
```
symphony-thibault, line 1: Not sure that we should put this service in a dedicated package. We can keep it as part of `com.symphony.bdk.core.service`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
...
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
+
+/**
+ * A class for implementing the user service.
```
symphony-thibault, line 32: Service class for managing users.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
...
+  public UserService(UserApi userApi, UsersApi usersApi, AuthSession authSession) {
+    this.userApi = userApi;
+    this.usersApi = usersApi;
+    this.authSession = authSession;
+  }
+
+  /**
+   * Retrieve user details of a particular user.
+   * @param uid User Id
+   * @return Details of the user.
```
symphony-thibault, line 62: I'll be a little bit boring :) But I do think that it would be really nice to provide the link to our API reference for each implemented endpoint. That would definitely help developers to get more information about an endpoint. I however agree that it is quite obvious for this one, but let's try to introduce this good practice in the BDK. 
```java
/**
  * [...]
  * @see <a href="https://developers.symphony.com/restapi/reference#get-user-v2">Get User v2</a>
  */
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  /**
+   * Retrieve a list of users in the company (pod) by a filter.
+   * If no filter is given, retrieving all users in the company (pod).
+   *
+   * @param filter using to filter users by
+   * @return List of retrieved users
+   */
+  public List<V2UserDetail> listUsersDetailByFilter(@Nullable UserFilter filter) {
```
symphony-thibault, line 79: Better to split this method in 2 different ones `listUsersDetail()` and `listUsersDetail(@Nonnull UserFilter filter)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
...
+      throw new ApiRuntimeException(apiException);
+    }
+  }
+
+  /**
+   * Remove a role from an user.
+   * @param uid User Id
+   * @param roleId Role Id
+   */
+  public void removeRoleFromUser(@NonNull Long uid, @NonNull String roleId) {
```
symphony-thibault, line 111: We need to provide an `enum` here containing available roles (cf. https://github.com/SymphonyOSF/SBE/blob/dev/persistence/data-common/src/main/java/com/symphony/data/common/roles/RoleId.java)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
...
+      throw new ApiRuntimeException(apiException);
+    }
+  }
+
+  /**
+   * Update avatar of an user
+   * @param uid User Id
+   * @param image The avatar image for the user profile picture.The image must be a base64-encoded.
+   */
+  public void updateAvatarOfUser(@NonNull Long uid, @NonNull String image) {
```
symphony-thibault, line 138: Could we also provide prototypes that accept `byte[]` or `InputStream` for the image parameter?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,394 @@
...
+  /**
+   * Get user by id
+   *
+   * @param uid User Id
+   * @param local If true then a local DB search will be performed and only local pod users will be
+   *             returned. If absent or false then a directory search will be performed and users
+   *             from other pods who are visible to the calling user will also be returned.
+   * @return User found by uid
+   */
+  public UserV2 getUserById(@NonNull Long uid, @Nullable Boolean local) {
```
symphony-thibault, line 284: For each method that accepts `local` as param: let's provide a second one without `local` (default value will be `false` according to our doc )

symphony-thibault:
Last minute idea: we could create a custom annotation `@OBOEnabled` that we would place on top of methods that actually accept OBO authentication. Not very useful at the moment since we don't support OBO for services yet, but that could be really interesting in the future, I think...

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/OboEnabled.java
```diff
@@ -0,0 +1,12 @@
+package com.symphony.bdk.core.service;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
```
symphony-thibault, line 8: Javadoc missing + add link to https://developers.symphony.com/restapi/docs/get-started-with-obo

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/constant/RoleId.java
```diff
@@ -0,0 +1,26 @@
+package com.symphony.bdk.core.service.user.constant;
+
+/**
+ * Static roles that have special logic attached to them.
```
symphony-thibault, line 4: + `@see` link to https://developers.symphony.com/restapi/docs/permissions

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,530 @@
...
+
+  public UserService(UserApi userApi, UsersApi usersApi, AuthSession authSession) {
+    this.userApi = userApi;
+    this.usersApi = usersApi;
+    this.authSession = authSession;
+  }
+
+  /**
+   * Retrieve user details of a particular user.
+   * @see <a href="https://developers.symphony.com/restapi/reference#get-user-v2">Get User v2</a>
```
symphony-thibault, line 66: `@see` is usually placed at the end of the Javadoc comment block (cf. https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html)

symphony-thibault (approved):
LGTM 👍

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,520 @@
...
+    }
+  }
+
+  /**
+   * Get user by id
+   *
+   * @param uid User Id
+   * @return    User found by uid
+   */
+  public UserV2 getUserById(@NonNull Long uid) {
```
symphony-youri, line 358: We could have a getUser(UserId id) method to cover both cases (long id and string email)

And then UserId.id(123) and UserId.email("bob@mail.com")
This way we don't mess up with primitive types and avoid having methods suffixed with byId/byEmail

But maybe the API does not always support id and email?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,520 @@
...
+  }
+
+  /**
+   * Update avatar of an user
+   *
+   * @param uid   User Id
+   * @param image The avatar image for the user profile picture.The image must be a base64-encoded.
+   * @see         <a href="https://developers.symphony.com/restapi/reference#update-user-avatar">Update User Avatar</a>
+   */
+  public void updateAvatarOfUser(@NonNull Long uid, @NonNull String image) {
```
symphony-youri, line 165: As a API user in Java I would perhaps prefer to send an inputstream here (and then have the SDK do the encoding for me)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -0,0 +1,520 @@
...
+  }
+
+  /**
+   * Update avatar of an user
+   *
+   * @param uid   User Id
+   * @param image The avatar image in bytes array for the user profile picture.
+   * @see         <a href="https://developers.symphony.com/restapi/reference#update-user-avatar">Update User Avatar</a>
+   */
+  public void updateAvatarOfUser(@NonNull Long uid, @NonNull byte[] image) {
```
symphony-youri, line 181: ah that one is better :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
symphony-hong: the API for getting user accept only one of the params: username, uid or email. If we pass more than one param, the error 400 is returned. That's why I splitted it to 3 different methods.

symphony-youri: It does not prevent the UserService to have the logic to set the username/uid/email based on the UserId passed as a parameter

but do **all** the APIs related to user support passing username, uid or email? (otherwise we would not have a consistent API where UserId can be used everywhere)

symphony-hong: Nope, only the API for getting user: 
https://github.com/symphonyoss/symphony-api-spec/blob/master/pod/pod-api-public.yaml#L1784
and the API for seaching users:
https://github.com/symphonyoss/symphony-api-spec/blob/master/pod/pod-api-public.yaml#L1426

PR #196, opened 2020-09-02
APP-3018: Using varargs in constructor and setAttachment methods

symphony-hong:


symphony-elias (approved):
LGTM

symphony-bdk-legacy/symphony-api-client-java/src/main/java/model/OutboundMessage.java
```diff
@@ -23,37 +24,25 @@ public OutboundMessage(String message, String data) {
         this.data = data;
     }
 
-    public OutboundMessage(String message, File[] attachment) {
+    public OutboundMessage(String message, File... attachment) {
```
symphony-thibault, line 27: What happen if we pass a `null` value here?

symphony-hong: We have to cast it to File class, then the attachment array will contain only 1 null element.

ystan- (approved):
LGTM

PR #210, opened 2020-09-11
Enhancement for MockApiClient

symphony-hong:
Throw MockApiClientException when calling wrong path or wrong method


symphony-thibault (approved):
LGTM 👍

PR #212, opened 2020-09-14
Speed up unittest by using MockApiClient

symphony-hong:
Update user service unittest using MockApiClient


symphony-elias (approved):
LGTM 
Unit tests are indeed much faster

symphony-thibault:
I much time did we save by using this new strategy? (just by curiosity 😄 )

symphony-hong:
> I much time did we save by using this new strategy? (just by curiosity 😄 )

In this PR, I updated only on UserService test and the BDK 2.0 check time reduced from more than 6 mins to 1m14s.

PR #213, opened 2020-09-14
Added new profile to exclude 2.0 examples when releasing

symphony-elias:
This is to exclude the examples module to be excluded when releasing BDK2.0 as we don't want to release them. When releasing BDK2.0, we will have to use both profiles `2.0-no-example` and `release`.

Alternatives using `skipNexusStagingDeployMojo` documented [here](https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deploymentl) seems cumbersome as the latest module in the reactor must have this configured in order to work.

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-deploy-plugin</artifactId>
    <version>2.7</version>
    <configuration>
        <skip>true</skip>
    </configuration>
  </plugin>
```
does not work as nexus plugin is used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -161,6 +161,20 @@
         </profile>
 
         <profile>
+            <id>2.0-no-example</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <modules>
+                <module>symphony-bdk-bom</module>
+                <module>symphony-bdk-core</module>
+                <module>symphony-bdk-core-invokers</module>
+                <module>symphony-bdk-spring</module>
+                <module>symphony-bdk-template</module>
+            </modules>
+        </profile>
+
+        <profile>
             <id>release</id>
             <activation>
                 <activeByDefault>false</activeByDefault>


pom.xml
```diff
@@ -160,6 +160,20 @@
             </modules>
         </profile>
 
+        <profile>
+            <id>2.0-no-example</id>
```
symphony-thibault, line 164: Question, do we still need to have this `2.0` profile?

symphony-hong: @symphony-thibault or directly exclude the example module out of profile `2.0`, then we don't need to update the CircleCI workflow

symphony-youri:
> This is to exclude the examples module to be excluded when releasing BDK2.0 as we don't want to release them. When releasing BDK2.0, we will have to use both profiles `2.0-no-example` and `release`.
> 
> Alternatives using `skipNexusStagingDeployMojo` documented [here](https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deploymentl) seems cumbersome as the latest module in the reactor must have this configured in order to work.


I did not get that point here, it would be nice to do it that way to make the build as simple as possible (knowing which profile to activate and when makes it complicate


> 
> ```
> <plugin>
>     <groupId>org.apache.maven.plugins</groupId>
>     <artifactId>maven-deploy-plugin</artifactId>
>     <version>2.7</version>
>     <configuration>
>         <skip>true</skip>
>     </configuration>
>   </plugin>
> ```
> 
> does not work as nexus plugin is used.

PR #215, opened 2020-09-14
APP-3056 Fix Javadoc generation

symphony-thibault:


symphony-youri (approved):
We could still use checkstyle for be more restrictive and enforce formatting
But I guess editorconfig is easy enough to configure to hope that every one is using (and formatting changed code before pushing)

symphony-hong (approved):
LGTM!

PR #218, opened 2020-09-15
APP-2846: Obo Certificate Authentication

symphony-hong:
Create AuthSessionOboCerImpl implements the AuthSession to represent
a authsession after a bot authenticate OBO user

OboAuthenticatorCertImpl implements the OboAuthenticator for handling
the Obo authentication stuff for a bot using Certificate

In ApiClientFactory, create a method ApiClientFactory#getExtAppSessionAuthClient
to get a dedicated sessionauth client for extension app authentication configured
with the keystore of the extension app and the password

In AuthenticatorFactory, check if the extension app is configured with
certificate or not. If yes, AuthenticatorFactory#getOboAuthenticator returns
an OboAuthenticatorCertImpl, otherwise, returns an OboAuthenticatorRsaImpl

Unittest added

Javadoc added


symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionOboCerImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void refresh() throws AuthUnauthorizedException {
+    if (this.userId != null) {
+      this.sessionToken = this.authenticator.retrieveOboSessionTokenByUserId(this.userId);
+    } else if (this.username != null) {
+      this.sessionToken = this.authenticator.retrieveOboSessionTokenByUsername(username);
```
symphony-thibault, line 63: `this.username` above so `this.username` as argument ;-)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionOboCerImpl.java
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+      throw new IllegalStateException("Both userId and username are null. One of them is mandatory to perform OBO authentication refresh.");
+    }
+  }
+
+  protected OboAuthenticator getAuthenticator() {
```
symphony-thibault, line 69: I'd add a Javadoc comment: "Visible for testing"

symphony-bdk-core/src/main/java/com/symphony/bdk/core/client/ApiClientFactory.java
```diff
@@ -95,6 +96,22 @@ public ApiClient getSessionAuthClient() {
     return buildClientWithCertificate(this.config.getSessionAuth().getBasePath() + "/sessionauth");
```
symphony-thibault, line 96: I'd change this line by: 
```java
return buildClientWithCertificate(this.config.getSessionAuth().getBasePath() + "/sessionauth", botConfig.getCertificatePath(), botConfig.getCertificatePassword());
```
and remove method `buildClientWithCertificate(String basePath)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkExtAppConfig.java
```diff
@@ -1,10 +1,10 @@
...
+import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
+
 import lombok.Getter;
 import lombok.Setter;
 
-import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
-
 @Getter
 @Setter
 public class BdkExtAppConfig {
```
symphony-thibault, line 10: `BdkExtAppConfig` looks like `BdkBotConfig` at some points, right?

symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/OboAuthenticatorCertImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final String APP_AUTHENTICATE = "/sessionauth/v1/app/authenticate";
+  private static final String APP_AUTHENTICATE_OBO_USERNAME = "/sessionauth/v1/app/username/{username}/authenticate";
+  private static final String APP_AUTHENTICATE_OBO_USERID = "/sessionauth/v1/app/user/{uid}/authenticate";
+
+  private OboAuthenticatorCertImpl authenticator;
+  private MockApiClient mockApiClient;
+
+  @BeforeEach
+  void init() {
+    mockApiClient = new MockApiClient();
```
symphony-thibault, line 27: If you use `this` for the `authenticator` then your should use `this` for the `mockApiClient` as well

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkExtAppConfig.java
symphony-hong: There is a small difference in field name between appId and username. For the others, it seems similar.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionOboCerImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import lombok.NonNull;
+import org.apiguardian.api.API;
+
+import javax.annotation.Nullable;
+
+/**
+ * {@link AuthSession} impl for OBO Certificate authentication mode.
+ */
+@API(status = API.Status.INTERNAL)
+public class AuthSessionOboCerImpl implements AuthSession {
```
symphony-elias, line 17: AuthSessionOboCerImpl or AuthSessionOboCer**t**Impl ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/client/ApiClientFactory.java
```diff
@@ -92,7 +93,30 @@ public ApiClient getAgentClient() {
...
+      throw new ApiClientInitializationException("For certificate authentication, " +
+          "certificatePath and certificatePassword must be set");
+    }
+
+    return buildClientWithCertificate(this.config.getSessionAuth().getBasePath() + "/sessionauth", botConfig.getCertificatePath(), botConfig.getCertificatePassword());
+  }
+
+  /**
+   * Returns a fully initialized {@link ApiClient} for the SessionAuth API using in Extension App Authentication.
+   * This only works with a extension app authenticate configured
```
symphony-elias, line 108: extension app authentication ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkExtAppConfig.java
symphony-elias: It could be great to have the common fields and common methods factorized in a base class, esp. `isCertificateAuthenticationConfigured()`, `certificatePath` and `certificatePassword`.
And it would enable you to extract the logic in ApiClientFactory lines 98 and 115

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/OboAuthenticatorCertImpl.java
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }
+  }
+
+  protected String retrieveOboSessionTokenByUsername(@NonNull String username) throws AuthUnauthorizedException {
+    final String appSessionToken = this.retrieveAppSessionToken();
+    try {
+      OboAuthResponse oboAuthResponse = this.authenticationApi.v1AppUsernameUsernameAuthenticatePost(username, appSessionToken);
+      return oboAuthResponse.getSessionToken();
+    } catch (ApiException e) {
```
symphony-elias, line 80: Could we find a way to extract this logic ? Only difference is the message in the `AuthUnauthorizedException`.

symphony-hong: I think we are fine with the current code. It's not a big block of code at all. 🤔

symphony-thibault (approved):
LGTM 👍

symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/OboAuthenticatorCertImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+        this.mockApiClient.getApiClient("/sessionauth")
+    );
+  }
+
+  @Test
+  void testAuthenticateByUsername() throws AuthUnauthorizedException {
+    this.mockApiClient.onPost(APP_AUTHENTICATE, "{ \"token\": \"1234\", \"name\": \"sessionToken\" }");
+    this.mockApiClient.onPost(APP_AUTHENTICATE_OBO_USERNAME.replace("{username}", "username"), "{ \"sessionToken\": \"1234\" }");
+    final AuthSession session = this.authenticator.authenticateByUsername("username");
+    assertNotNull(session);
```
symphony-youri, line 39: IMHO the checks on session is not required as we do have asserts on session after (so they would fail with NPE)
I would also recommend using line breaks to split different phases of the test, i.e Given/When/Then (the setup, what is tested and the checks)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkBotConfig.java
```diff
@@ -2,22 +2,11 @@
...
-import org.apache.commons.lang3.StringUtils;
 
 @Getter
 @Setter
-public class BdkBotConfig {
+public class BdkBotConfig extends BdkAuthenticationConfig {
 
     private String username;
-    private String privateKeyPath;
-    private String certificatePath;
```
symphony-youri, line 13: in term of configuration should'nt we have something like:

username: "mybot"
certificate:
  path:
  password

repeating the same prefix is often the sign that a sub structure should be used

but then to be consistent you would wonder if privateKeyPath should a privateKey item with a path sub item....
could we have other settings for a private key?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionOboCertImpl.java
```diff
@@ -0,0 +1,75 @@
+package com.symphony.bdk.core.auth.impl;
+
+import com.symphony.bdk.core.auth.AuthSession;
+
+import com.symphony.bdk.core.auth.OboAuthenticator;
+import com.symphony.bdk.core.auth.exception.AuthUnauthorizedException;
+
+import lombok.NonNull;
```
symphony-youri, line 8: do we have that NonNull annotation on purpose (vs the javax one)?
if so maybe we should explicit when we use that one vs the other

symphony-thibault: Import issue I guess, should be `javax.annotation.Nonnull`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkAuthenticationConfig.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class BdkAuthenticationConfig {
+
+  protected String privateKeyPath;
+  protected String certificatePath;
+  protected String certificatePassword;
```
symphony-thibault, line 14: Following @symphony-youri's comment below, I also think that we should change our configuration model to: 
```yaml
bot:
    username: 'my-bot'
    rsa:
        privateKeyPath: 
        privateKeyContent: 
    certificate:
        path: 
        password: 
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AuthSessionOboCertImpl.java
symphony-youri: NonNull from lombok actually helps us validate the configuration which could be useful here
(Although we perhaps want to provider our own validation for more complex use cases and checks or use javax validation)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkAuthenticationConfig.java
symphony-youri: but then should it be:

```
bot:
  username: 'my-bot'
  rsa:
    privateKey:
       path: / or content: 
  certificate:
    path: 
    password: 
```

😅

or maybe just rsa.path / rsa.content ? (and by the way is it really rsa or the notion of public/private keys? might be another algorithm)

so maybe just privateKey.path / privateKey.content?

symphony-youri (approved):

PR #223, opened 2020-09-16
SDK-136 Removed @Ignore annotation from class DatafeedClientV2Test

Chaoui13:


Chaoui13:
In order to the method readDatafeedConnectionTimeout() succeeds, I had to overload the method stubPost() of the class BotTest by adding the delay in the parameters

PR #227, opened 2020-09-16
Sdk-133 Create test coverage for the class clients.symphony.api.HealthcheckClient

Haiddi-Avencod:


pom.xml
```diff
@@ -219,4 +219,13 @@
...
     </profiles>
+
+    <dependencies>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
```
symphony-thibault, line 230: We don't need to add any dependency in the root `pom.xml`. Do you know why you needed to add here?

Chaoui13: Because I wanted to execute all the tests by the contextual menu of the parent > "Run all tests" and I had an error message that said that junit.rar was missing. I searched on Google and found this solution. After that I ran the tests without any problem. Isn't it a good solution ?

symphony-thibault: The best solution to execute all tests is to run Maven command : `mvn clean package`. The root `pom.xml` must not contain any dependency. 

You should execute tests from module (`symphony-api-client-java`) only

Chaoui13: Should I make a ticket, assign it to me and resolve it ?

symphony-thibault: No it's ok @Haiddi-Avencod, you should just revert this change in root `pom.xml` and run `mvn clean package` from your terminal when you want to execute all tests

Haiddi-Avencod: Here's the result when I run `mvn clean package`
mvn_clean_package

symphony-bdk-legacy/symphony-api-client-java/src/test/java/it/clients/symphony/api/HealthcheckClientTest.java
```diff
@@ -35,10 +34,24 @@ public void performHealthCheckSuccess() {
...
+
+      assertTrue(response.getPodConnectivity());
+      assertTrue(response.getKeyManagerConnectivity());
+      assertTrue(response.getEncryptDecryptSuccess());
+      assertEquals("1.54.1", response.getPodVersion());
+      assertEquals("2.54.0", response.getAgentVersion());
+      assertTrue(response.getAgentServiceUser());
+      assertTrue(response.getCeServiceUser());
+
+    } catch (Exception e) {
```
symphony-youri, line 52: An exception raised during a test will fail it, so we don't need to try/catch them

Haiddi-Avencod: Ok thanks, I will fix it.

Otherwise in the method **performHealthCheckSuccessWithUnknownField()** I don't know how to get the value of the field **unknownField**. I didn't find a getter for this field in the response object.

pom.xml
Haiddi-Avencod: I fixed it ;)

PR #232, opened 2020-09-17
APP-2821: Implement a configurable retry mechanism

symphony-elias:
### Ticket
[APP-2821](https://perzoinc.atlassian.net/browse/APP-2821)

### Description
Goal is to provide a generic retry mechanism for http calls. Refactored DFv1 and DFv2. Will extend this to all services.
Also contains code reformatting (according to the .editorconfig file : 2 spaces for indentation).

### Dependencies
None

### Checklist


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -26,93 +32,91 @@
...
 
-    private boolean isSelfGeneratedEvent(V4Event event) {
-        return event.getInitiator() != null && event.getInitiator().getUser() != null
-                && event.getInitiator().getUser().getUsername() != null
-                && event.getInitiator().getUser().getUsername().equals(this.bdkConfig.getBot().getUsername());
-    }
+  private boolean isSelfGeneratedEvent(V4Event event) {
+    return event.getInitiator() != null && event.getInitiator().getUser() != null
+        && event.getInitiator().getUser().getUsername() != null
+        && event.getInitiator().getUser().getUsername().equals(this.bdkConfig.getBot().getUsername());
```
symphony-thibault, line 90: Let's clean this code a bit (even if you are not author). We can here simply `return event.getInitiator().getUser().getUsername().equals(this.bdkConfig.getBot().getUsername())` as sub-objects cannot be `null`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
```diff
@@ -26,93 +32,91 @@
...
-            log.info("Retry in {} secs...", interval);
-        });
-        return retry;
+  protected boolean isNetworkOrServerOrUnauthorizedOrClientError(Throwable t) {
+    if (t instanceof ApiException && t.getSuppressed().length == 0) {
+      ApiException apiException = (ApiException) t;
+      return apiException.isServerError() || apiException.isUnauthorized() || apiException.isClientError();
     }
+    return t instanceof ProcessingException;
+  }
```
symphony-thibault, line 103: I fill like we could move so a separate class, as it is not only related to Datafeed

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/RetryWithRecovery.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Map;
+import java.util.function.Predicate;
+
+/**
+ * This class aims to implement a retry mechanism (on top of a{@link Retry})
+ * with different recovery strategies based on predicates.
+ * @param <T> the type of the object returned by {@link #execute()}
+ */
+@Slf4j
+public class RetryWithRecovery<T> {
```
symphony-thibault, line 19: I think that we should try to separate contract from implementation with this feature. Do we want want to make it available for developers? Are we sure that we will keep `resilience4j` as retry library in the future?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/RetryWithRecovery.java
```diff
@@ -0,0 +1,119 @@
...
+   * Constructor with no predicate on when to ignore an {@link ApiException},
+   * i.e. ApiExceptions will never be ignored.
+   * @param name the name of the {@link Retry} service.
+   * @param bdkRetryConfig the retry configuration to be used.
+   * @param supplier the supplier responsible to provide the object of param type T and which may throw an {@link ApiException}.
+   * @param retryOnExceptionPredicate predicate on a thrown {@link ApiException} to know if call should be retried.
+   * @param recoveryStrategies mapping between {@link Predicate<ApiException>} and the corresponding recovery functions to be executed before retrying.
+   *                           If several predicates match, all corresponding consumers will be executed.
+   */
+  public RetryWithRecovery(String name, BdkRetryConfig bdkRetryConfig, SupplierWithApiException<T> supplier,
```
symphony-thibault, line 35: I'd prefer to have `RetryFactory` that already hold the `BdkRetryConfig`. This would avoid passing it `RetryWithRecovery` each time we create a new instance

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ConsumerWithThrowable.java
```diff
@@ -0,0 +1,12 @@
+package com.symphony.bdk.core.util;
```
symphony-thibault, line 1: Let's move this class (and sisters) to package `com.symphony.bdk.core.util.function`, following [`java.util.function`](https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html#:~:text=Represents%20an%20operation%20that%20accepts%20two%20input%20arguments%20and%20returns%20no%20result.&text=Represents%20a%20function%20that%20accepts%20two%20arguments%20and%20produces%20a%20result.&text=Represents%20an%20operation%20upon%20two,same%20type%20as%20the%20operands.)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ConsumerWithThrowable.java
```diff
@@ -0,0 +1,12 @@
...
+import com.symphony.bdk.core.api.invoker.ApiException;
+
+/**
+ * Functional interface which consumes an {@link ApiException} and may throw a {@link Throwable}.
+ * This is used to specify recovery functions in {@link ConsumerWithThrowable}.
+ */
+@FunctionalInterface
+public interface ConsumerWithThrowable {
+  void consume(ApiException e) throws Throwable;
+}
```
symphony-thibault, line 12: Could we make it more generic? 

```java
public interface ConsumerWithThrowable<E extends Throwable> {

  void consume(E e) throws Throwable;
}
```

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiException.java
```diff
@@ -75,11 +77,15 @@ public boolean isClientError() {
...
     /**
-     * Check if response status is server error or not
+     * Check if response status is server error and not an internal server error
      *
-     * @return true if response status equals or greater than 500, false otherwise
+     * @return true if response status strictly greater than 500, false otherwise
      */
     public boolean isServerError() {
-        return this.code >= HttpURLConnection.HTTP_INTERNAL_ERROR;
+        return this.code > HttpURLConnection.HTTP_INTERNAL_ERROR;
```
symphony-hong, line 85: this means we don't care about the `HTTP_INTERNAL_ERROR`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ConsumerWithThrowable.java
symphony-elias: Yes we could but if we want to have a fully generic retry mechanism, we will have to add a type parameter for the exception in the RetryWithRecovery class. We cannot catch a type parameter in a try/catch clause. And instanceof works bad with parametrized type : https://stackoverflow.com/questions/1570073/java-instanceof-and-generics

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiException.java
symphony-elias: Yes indeed, in this case we assume it is not worth retrying. This was agreed with @symphony-thibault.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/function/RetryWithRecovery.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.symphony.bdk.core.config.model.BdkRetryConfig;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Map;
+import java.util.function.Predicate;
+
+@Slf4j
+public abstract class RetryWithRecovery<T> {
```
symphony-thibault, line 16: Create a package `com.symphony.bdk.core.retry` and move this class there

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/function/RetryWithRecoveryBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+import com.symphony.bdk.core.api.invoker.ApiException;
+import com.symphony.bdk.core.config.model.BdkRetryConfig;
+
+import javax.ws.rs.ProcessingException;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Predicate;
+
+public class RetryWithRecoveryBuilder<T> {
```
symphony-thibault, line 12: Create a package `com.symphony.bdk.core.retry` and move this class there

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/function/Resilience4jRetryWithRecovery.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Map;
+import java.util.function.Predicate;
+
+/**
+ * This class aims to implement a retry mechanism (on top of a{@link Retry})
+ * with different recovery strategies based on predicates.
+ * @param <T> the type of the object returned by {@link #execute()}
+ */
+@Slf4j
+public class Resilience4jRetryWithRecovery<T> extends RetryWithRecovery<T> {
```
symphony-thibault, line 21: Create a package `com.symphony.bdk.core.retry.resilience4j` and move this class there

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/SessionService.java
```diff
@@ -25,10 +28,7 @@
    * @return Bot session info.
    */
   public UserV2 getSession(AuthSession authSession) {
-    try {
-      return this.sessionApi.v2SessioninfoGet(authSession.getSessionToken()) ;
-    } catch (ApiException ex) {
-      throw new ApiRuntimeException(ex);
-    }
+    return RetryWithRecovery.executeAndRetry("getSession",
```
symphony-thibault, line 31: Passing the `BdkRetryConfig` is a bit repetitive, could you modify your design to avoid this? 

I remember that we talked about a `RetryFactory` or equivalent that would have been initialized at `SymphonyBdk` level, including configuration

symphony-elias: Yes but we still have to pass something to the classes, either a `BdkRetryConfig` or a `RetryWithRecoveryBuilder` initialised with the config. In either cases, we have to pass something to every classes.
Or maybe I didn't get something in your proposal.

symphony-thibault: Correct, we would still have to pass something to services. But IMHO the code would be more functional if we pass "a retry thing" rather than "a retry config"

symphony-bdk-core-invokers/symphony-bdk-core-invoker-api/src/main/java/com/symphony/bdk/core/api/invoker/ApiException.java
```diff
@@ -3,83 +3,104 @@
...
  * Initially generated by the OpenAPI Maven Generator
  */
 @Getter
 @API(status = API.Status.STABLE)
 public class ApiException extends Exception {
 
-    private int code = 0;
-    private Map<String, List<String>> responseHeaders = null;
-    private String responseBody = null;
+  private int code = 0;
```
symphony-thibault, line 23: Be careful as @symphony-hong is also changing the name of this module in `//`

symphony-elias: Space changes removed

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/MessageService.java
```diff
@@ -357,11 +369,15 @@ public MessageReceiptDetailResponse listMessageReceipts(@Nonnull String messageI
...
+    return executeAndRetry("getMessageRelationships", () -> defaultApi.v1AdminMessagesMessageIdMetadataRelationshipsGet(
         authSession.getSessionToken(), ApiUtils.getUserAgent(), messageId));
   }
 
   private static Long getEpochMillis(Instant instant) {
     return instant == null ? null : instant.toEpochMilli();
   }
+
+  private <T> T executeAndRetry(String name, SupplierWithApiException<T> supplier) {
+    return RetryWithRecovery.executeAndRetry(retryBuilder, name, supplier);
```
symphony-thibault, line 381: Could we have `this.retryBuilder.execute(name, supplier)`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/exception/NestedRetryException.java
```diff
@@ -0,0 +1,13 @@
+package com.symphony.bdk.core.service.datafeed.exception;
+
+import com.symphony.bdk.core.retry.RetryWithRecovery;
+
+/**
+ * Exception thrown when recovery strategy in a {@link RetryWithRecovery} failed.
+ * Especially used in DataFeed services.
```
symphony-thibault, line 7: `Especially used by {@link DatafeedService} implementations.`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/retry/RetryWithRecoveryBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+ * @param <T> the type to be returned by {@link RetryWithRecovery#execute()}.
+ */
+@API(status = API.Status.INTERNAL)
+public class RetryWithRecoveryBuilder<T> {
+  private String name;
+  private BdkRetryConfig retryConfig;
+  private SupplierWithApiException<T> supplier;
+  private Predicate<Throwable> retryOnExceptionPredicate;
+  private Predicate<ApiException> ignoreException;
+  private Map<Predicate<ApiException>, ConsumerWithThrowable> recoveryStrategies;
```
symphony-thibault, line 29: What if we want to generalize this API to something else than our BDK usage? Having `ApiException` really stick the design to our usage

symphony-elias: Yes I agree, but this works because we `catch(ApiException e)`. If we have a param type E, we cannot `catch(E e)` anymore as it is forbidden by java. We could try to `catch(Throwable t)` and check `if (t instanceof E)` but this is not possible either. We could circumvent it but it makes things quite cumbersome : see https://stackoverflow.com/questions/1570073/java-instanceof-and-generics

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/AbstractDatafeedService.java
symphony-elias: Indeed but it can be null in unit tests.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/MessageService.java
symphony-elias: I just realized it won't work because we use a same base `retryBuilder` for suppliers returning different types. So we have to keep it static with a param type which can be specified

symphony-thibault: Understood 👍

symphony-youri:
Looks good to me
Perhaps we can also try to run the datafeed locally and try to break in different ways (break auth, cut wifi, shutdown server, etc) and see how it copes and if it keeps running as expected, maybe this manual testing will expose cases we did not think of (and also see if error reporting is good enough to help understand the error_

symphony-thibault (approved):
LGTM 👍

symphony-youri (approved):

PR #234, opened 2020-09-21
APP-3032: Revamping of the API Invoker

symphony-hong:
### Ticket
[3032](https://perzoinc.atlassian.net/browse/APP-3032)

### Description
BDK: Revamping of the API Invoker

### Dependencies

### Checklist


symphony-bdk-core/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation 'org.mapstruct:mapstruct:1.3.1.Final'
     annotationProcessor 'org.mapstruct:mapstruct-processor:1.3.1.Final'
 
-    implementation 'org.apiguardian:apiguardian-api'
+    compile 'org.apiguardian:apiguardian-api'
```
symphony-youri, line 40: https://docs.gradle.org/current/userguide/java_library_plugin.html 

_The compile configuration still exists but should not be used as it will not offer the guarantees that the api and implementation configurations provide._

Did you have errors with the implementation scope?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.ws.rs.core.GenericType;
```
symphony-youri, line 21: That should go away at some point right? If we want to be independant?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
```diff
@@ -0,0 +1,150 @@
...
+
+@Slf4j
+public class HttpClient {
```
symphony-youri, line 24: Our http client looks very similar to the jersey one, will this api work with other clients? (we could try openfeign and the jdk ones for instance)

symphony-youri (commented):
If I understand correctly the goal of the http modules is to be able to choose the http client implementation you want to use, do we have reasons to offer that possibility vs just choosing one and sticking to it?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
```diff
@@ -0,0 +1,150 @@
+package com.symphony.bdk.core;
```
symphony-thibault, line 1: This class should be located in the `http` module as well. It could be used outside of the `core`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
symphony-thibault: And Spring WebClient (priority for SpringBoot integration)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
```diff
@@ -0,0 +1,150 @@
...
+      this.keyStorePassword = keyStorePassword;
+      return this;
+    }
+
+    public Builder trustStore(String trustStorePath, String trustStorePassword) {
+      this.trustStorePath = trustStorePath;
+      this.trustStorePassword = trustStorePassword;
+      return this;
+    }
+
```
symphony-thibault, line 111: keyStore and trustStore should both accept `byte[]`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ProviderLoader.java
```diff
@@ -0,0 +1,43 @@
...
+@Slf4j
+@Generated
+public class ProviderLoader {
+
+  /**
+   * Load {@link ApiClient} implementation class using {@link ServiceLoader}.
+   *
+   * @return an {@link ApiClientBuilderProvider}.
+   */
+  public static ApiClientBuilderProvider findApiClientBuilderProvider() {
```
symphony-thibault, line 25: Now you refactored it as a utility class, could we try to  make it a bit more generic? `public static <T> T lookupSingleService(Class<T> clz)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ProviderLoader.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Generated;
+import lombok.extern.slf4j.Slf4j;
+import org.apiguardian.api.API;
+
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+@API(status = API.Status.INTERNAL)
```
symphony-thibault, line 15: This annotation goes just above the class name

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
symphony-thibault: That would be super nice indeed if we could get rif of this classs, maybe by using our own generic type?

symphony-bdk-examples/bdk-core-examples/build.gradle
```diff
@@ -16,5 +17,7 @@ dependencies {
 
     testCompileOnly 'org.projectlombok:lombok'
     testAnnotationProcessor 'org.projectlombok:lombok'
+
+    implementation 'org.glassfish.jersey.media:jersey-media-json-jackson'
```
symphony-thibault, line 21: Should be transitive through `symphony-bdk-http-jersey2`. Can we do that @symphony-youri?

symphony-bdk-core/build.gradle
symphony-hong: when I built the project, there are some warning about the enum `API.Status`. I had a look at the api-guardian [Readme](https://github.com/apiguardian-team/apiguardian), and changed to compile scope and the warning disappear.

symphony-bdk-examples/bdk-core-examples/build.gradle
symphony-youri: To me we should do 2 things here:
- implementation 'org.glassfish.jersey.core:jersey-client' in the example module because in HttpClientExampleMain we import the generic type
- in bdk-core change implementation 'org.glassfish.jersey.core:jersey-client' to api and add the plugin java-library since we are leaking the GenericType in the HttpClient class

symphony-bdk-core/build.gradle
symphony-youri: ah then we can add the java-library plugin in bdk-core and change it to 

    api 'org.apiguardian:apiguardian-api'
and we can do the same for jackson there are warnings for that too

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
symphony-hong: But the problem is that the generic `ApiClient#invokeAPI` used this `javax.ws.rs.core.GenericType`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/HttpClient.java
symphony-elias: in scope of https://perzoinc.atlassian.net/browse/APP-3069

symphony-thibault: RSA private and key/trustStore are different things

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ProviderLoader.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+@Slf4j
+@Generated
+@API(status = API.Status.INTERNAL)
+public class ProviderLoader {
```
symphony-thibault, line 15: We could call this class `ServiceLookup` as it can load any kind of service. 

(I now remember my old time at school when we were developing [Netbeans plugins](http://bits.netbeans.org/dev/javadoc/org-openide-util-lookup/org/openide/util/Lookup.html))

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ServiceLookup.java
```diff
@@ -0,0 +1,38 @@
...
+    final ServiceLoader<T> classServiceLoader = ServiceLoader.load(clz);
+
+    final List<T> services = StreamSupport.stream(classServiceLoader.spliterator(), false)
+            .collect(Collectors.toList());
+
+    if (services.isEmpty()) {
+      throw new IllegalStateException("No service implementation found in classpath.");
+    } else if (services.size() > 1) {
+      log.warn("More than 1 service implementation found in classpath, will use : {}",
+          services.stream().findFirst().get());
```
symphony-youri, line 33: just services.get(0) ? (would avoid the warning about optional)

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/HttpClientExampleMain.java
```diff
@@ -0,0 +1,45 @@
...
+    BdkConfig config = loadFromSymphonyDir("config.yaml");
+    final SymphonyBdk bdk = new SymphonyBdk(config);
+    PrivateKey privateKey = jwtHelper.parseRsaPrivateKey(
+        IOUtils.toString(new FileInputStream(config.getBot().getPrivateKeyPath()), StandardCharsets.UTF_8));
+    final String jwt =
+        jwtHelper.createSignedJwt(config.getBot().getUsername(), JwtHelper.JWT_EXPIRATION_MILLIS, privateKey);
+    final AuthenticateRequest req = new AuthenticateRequest();
+    req.setToken(jwt);
+
+    Token token = bdk.http()
```
symphony-youri, line 39: if the http client is meant to be used for other usecases (outside of the SDK), perhaps we don't need to make it accessible in the BDK object (you would just create the Http client using the builder on its own)

symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/HttpClient.java
```diff
@@ -0,0 +1,241 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * HttpClient is the main entry point to fluent API used to build an {@link ApiClient} and execute
+ * client requests in order to consume responses returned
+ */
+public class HttpClient {
```
symphony-youri, line 14: if we want to enforce the use of the builder we can add a private constructor here

symphony-youri (approved):
you should be able to remove the dependency on jersey in the http api module

added a few small comments, otherwise it looks good to me, we can merge that one and then move one with the webclient in a separate PR if you prefer

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/HttpClientExampleMain.java
symphony-hong: I only added the constructor for unittest purpose. Jacoco complained that the code coverage on `HttpClient` didn't reach 90% even when intellij said 100%, I think it's because the default constructor is not tested (not really sure about it)

symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/HttpClient.java
symphony-hong: I only added the constructor for unittest purpose. Jacoco complained that the code coverage on HttpClient didn't reach 90% even when intellij said 100%, I think it's because the default constructor is not tested (not really sure about it)

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/HttpClientExampleMain.java
symphony-hong: this idea is suggested by @symphony-thibault to make the bdk usage more fluent. @symphony-thibault WDYT ?

symphony-thibault: This way developers don't need to pass the `ApiClientBuilderProvider`, it uses the one defined in `SymphonyBdk` found from ServiceLoader

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/HttpClientExampleMain.java
```diff
@@ -0,0 +1,45 @@
...
+        .path("https://devx1.symphony.com/login/pubkey/authenticate")
+        .body(req)
+        .post(new GenericClass<Token>() {});
+    System.out.println(token);
```
symphony-thibault, line 43: grrrrr :)

symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/util/GenericClass.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Represents a generic response entity type {@code T}.
+ * Supports in-line instantiation of objects that represent generic types with
+ * actual type parameters. An object that represents any parameterized type may
+ * be obtained by sub-classing {@code GenericClass}.
+ *
+ * @param <T> the generic type parameter.
+ */
+public abstract class GenericClass<T> {
```
symphony-thibault, line 14: `GenericType<T>`? Or why not just `Generic<T>`?

symphony-thibault (approved):
LGTM 👍

PR #245, opened 2020-09-28
Jacoco reports generated from Gradle build

symphony-thibault:
Jacoco reports were not generated from Gradle build. This PR fixes that.


symphony-elias (approved):
LGTM

PR #246, opened 2020-09-28
APP-2999: Updated jenkinsfile for uploading snapshots and releases

symphony-elias:
### Ticket
[APP-2999](https://perzoinc.atlassian.net/browse/APP-2999)

### Description
Updated jenkinsfile for releasing

Uploading to symphony repo works fine but uploading to nexus repository does not work as signing only possible with keyring files, see https://github.com/bmuschko/gradle-nexus-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/nexus/NexusPlugin.groovy#L101. 
Opened https://perzoinc.atlassian.net/browse/WAR-4250 to have the keyring file uploaded.

Successful build : https://warpdrive-lab.dev.symphony.com/jenkins/view/DevX-Basilic/job/bdk-deploy-elias/14/console
With no PARAM_RELEASE_VERSION: https://warpdrive-lab.dev.symphony.com/jenkins/view/DevX-Basilic/job/bdk-nexus-elias/16/console (closeRepository fails because -SNAPSHOT versions are put in the snapshot repo of nexus, not in the staging one)
With a PARAM_RELEASE_VERSION: https://warpdrive-lab.dev.symphony.com/jenkins/view/DevX-Basilic/job/bdk-nexus-elias/11/console

### Checklist


jenkins/Jenkinsfile-publish
```diff
@@ -0,0 +1,36 @@
+// used by https://warpdrive-lab.dev.symphony.com/jenkins/view/DevX-Basilic/job/bdk-deploy-java/
```
symphony-youri, line 1: maybe we don't want to leak such internal URL (well too late :))

jenkins/Jenkinsfile-publish
```diff
@@ -0,0 +1,34 @@
+@Library(['sym-pipeline']) _
+
+import com.symphony.cicd.SymphonyCICDUtils
+
+cicdUtils = new SymphonyCICDUtils()
```
symphony-youri, line 5: I guess we could get rid of that variable too as it does not seem to be used

symphony-youri (approved):

jenkins/Jenkinsfile-publish
symphony-thibault: And it is an internal library, better if we avoid to share it on public repos

jenkins/Jenkinsfile-publish
```diff
@@ -0,0 +1,34 @@
...
+        "GIT_ORG=$PARAM_REPO_OWNER", 
+        "GIT_REPO=symphony-api-client-java"
+      ]) {
+
+        stage('Git Checkout') {
+            gitCheckout()
+        }
+
+        stage('Build project') {
+            sh './gradlew clean build'
```
symphony-thibault, line 26: `sh './gradlew'` ?

symphony-youri: while it will be the same to me it is ok for CI to make it more explicit
(adding the default task in the build.gradle is more about making it easy for the developer)

symphony-thibault (approved):
LGTM 👍

PR #248, opened 2020-09-28
APP-3077 Cleaned Maven files

symphony-thibault:
### Ticket
[APP-3077](APP-3077)

### Description
Removed Maven files from project.


symphony-youri (approved):
We can also remove target/ from .gitignore

PR #254, opened 2020-09-29
[POC] OBO services

symphony-thibault:
Example: 
```java
  public static void main(String[] args) throws BdkConfigException, AuthInitializationException, AuthUnauthorizedException {

    final SymphonyBdk bdk = SymphonyBdk.fromSymphonyDir("config.yaml");

    final AuthSession oboSession = bdk.obo("thibault.pensec");

    final V4Message message = bdk.messages().obo(oboSession).send(STREAM, MESSAGE);
    log.info("Message {} sent using OBO auth mode", message.getMessageId());
  }
```

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/MessageService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/MessageService.java
@@ -2,12 +2,10 @@ package com.symphony.bdk.core.service;
 
 
 import com.symphony.bdk.core.auth.AuthSession;
-import com.symphony.bdk.core.retry.RetryWithRecovery;
 import com.symphony.bdk.core.retry.RetryWithRecoveryBuilder;
 import com.symphony.bdk.core.service.pagination.PaginatedApi;
 import com.symphony.bdk.core.service.pagination.PaginatedService;
 import com.symphony.bdk.core.service.stream.constant.AttachmentSort;
-import com.symphony.bdk.core.util.function.SupplierWithApiException;
 import com.symphony.bdk.gen.api.AttachmentsApi;
 import com.symphony.bdk.gen.api.DefaultApi;
 import com.symphony.bdk.gen.api.MessageApi;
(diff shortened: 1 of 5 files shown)


symphony-youri:
Should we have an OboAuthSession that would be returned by bdk.obo()?
This way bdk.messages().obo() only accepts an OboAuthSession making it impossible to mistakenly pass another session like a bot session

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/OboExampleMain.java
```diff
@@ -0,0 +1,28 @@
...
+  private static final String STREAM = "2IFEMquh3pOHAxcgLF8jU3___ozwgwIVdA";
+  private static final String MESSAGE = "<messageML>Hello, World!</messageML>";
+
+  public static void main(String[] args) throws BdkConfigException, AuthInitializationException, AuthUnauthorizedException {
+
+    final SymphonyBdk bdk = SymphonyBdk.fromSymphonyDir("config.yaml");
+
+    final AuthSession oboSession = bdk.obo("thibault.pensec");
+
+    final V4Message message = bdk.messages().obo(oboSession).send(STREAM, MESSAGE);
```
symphony-youri, line 25: What about bdk.obo(oboSession).messages()

That you first tell who you are and then do something

It would be more consistent to have bdk.obo().messages() then bdk.obo().streams()...
This way you can discover the services available via OBO using autocomplete

But then we already have a method obo that is used to create the auth session, as a user of the BDK should I care about the auth session or should the BDK handle it for me?

it could be `bdk.obo("thibault.pensec").messages().send(...)`
and bdk.messages() is actually a shortcut for bdk.bot().messages()

You might also wonder if the BDK should authenticate when created or only on first use

Looking at the auth code, the AuthSession is not thread safe (but I'm not sure it is a problem we might just ending up doing multiple refresh at the same time but if we end up with valid tokens in the end it could be fine)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/MessageService.java
```diff
@@ -45,46 +42,72 @@
...
     this.attachmentsApi = attachmentsApi;
     this.authSession = authSession;
-    this.templateResolver = templateResolver;
     this.defaultApi = defaultApi;
-    this.retryBuilder = retryBuilder;
+  }
+
+  @Override
+  public OboMessageService obo(AuthSession oboSession) {
+    return new OboMessageService(this.messagesApi, oboSession, this.retryBuilder, templateResolver);
```
symphony-hong, line 110: new instance of `OboMessageService` is returned each time `obo(oboSession)` is called. I guess we’ll back to the memory problem.

PR #260, opened 2020-10-04
APP-3110 Handlebars template implementation

symphony-thibault:
### Ticket
[APP-3110](https://perzoinc.atlassian.net/browse/APP-3110)

### Description
[Handlerbars](https://github.com/jknack/handlebars.java) implementation for the Template API for adoption purpose, as it is the default implementation provided by legacy [sms-sdk-renderer-java](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-legacy/sms-sdk-renderer-java/README.md) library. 

I also decided to simplify this API by removing the built-in templates feature that is not very useful at the moment, as we are not really providing any real built-in template. We might reconsider this feature in the future, depending on developers needs for that, but for now I really think that we don't necessary need to publish it in order to avoid any question like "why it exists but there is only 1 `simpleMML` template?".

### Checklist


symphony-bdk-template/symphony-bdk-template-freemarker/src/main/java/com/symphony/bdk/template/freemarker/FreeMarkerEngine.java
```diff
@@ -8,53 +8,31 @@
...
   /**
    * {@inheritDoc}
    */
   @Override
   public Template newTemplateFromFile(String templatePath) throws TemplateException {
     try {
-      return getTemplateFromFile(templatePath);
+      final String directory = FilenameUtils.getFullPathNoEndSeparator(templatePath);
+      final String file = FilenameUtils.getName(templatePath);
+      final Configuration configuration = createConfiguration(); // for thread-safety, we need to re-create configuration
```
symphony-elias, line 33: Thanks for putting this useful, because FreeMarker does not recommend to create a new configuration each time we load a template.

symphony-elias (approved):
LGTM

docs/message.md
```diff
@@ -17,54 +17,73 @@ More precisely:
...
+engine implementations: 
+- [FreeMarker](https://freemarker.apache.org/) (through dependency `com.symphony.platformsolutions:symphony-bdk-template-freemarker`)
+- [Handlebars](https://github.com/jknack/handlebars.java) (through dependency `com.symphony.platformsolutions:symphony-bdk-template-handlebars`)
 
-## Send messages with templates
-The Message service also allows you to send messages using templates. So far, we only support [FreeMarker templates](https://freemarker.apache.org/),
-but we may add support for other template engines.
-For instance, if you have the following template file:
+### How to send a message from a template
+> In the code examples below, we will assume that FreeMarker as been selected as template engine implementation
```
symphony-youri, line 42: should we detail how the selection is made (i.e whatever is in the classpath is used)?
One could think that there is an API to call to select the engine

symphony-youri (approved):

symphony-bdk-template/symphony-bdk-template-freemarker/src/main/java/com/symphony/bdk/template/freemarker/FreeMarkerEngine.java
symphony-thibault: The risk here was to update static configuration at method call level, it could be problematic in multi-threaded environment such as Spring Boot applications.

docs/message.md
symphony-thibault: I documented how to select the implementation. And I also created `APP-3113` to fix the current behaviour which was to randomly decide which implementation is selected by default

PR #277, opened 2020-10-07
APP-3073: Enhancements on MessageService

symphony-hong:
### Ticket
[APP-3073](https://perzoinc.atlassian.net/browse/APP-3073)

### Description
Enhancements on message service 
Move MessageService to package com.symphony.bdk.core.service.message

Make the TemplateException extends RuntimeException -> Remove all the throws TemplateException.

Rename MessageService#send(streamId, template, Object) -> MessageService#sendWithTemplate
Provide MessageService#send(streamId, template) to send a templated message without passing any parameter instead of passing null to the existing method.

Provide method MessageService#send(streamId, message, data, attachment) to send a message with data and attachment (streamId can be replaced by a V4Stream instance)

Small fix on webclient api to perform multipart/form-data request.

Unittest added

### Dependencies

### Checklist


symphony-bdk-core/src/main/java/com/symphony/bdk/core/ServiceFactory.java
```diff
@@ -4,7 +4,7 @@
 import com.symphony.bdk.core.client.ApiClientFactory;
 import com.symphony.bdk.core.config.model.BdkConfig;
 import com.symphony.bdk.core.retry.RetryWithRecoveryBuilder;
-import com.symphony.bdk.core.service.MessageService;
+import com.symphony.bdk.core.service.message.MessageService;
```
symphony-thibault, line 7: This breaking change might impact our documentation even if I don't think that we specify packages, please take a look just in case. 

But it will for sure break `generator-symphony`: 
- https://github.com/SymphonyPlatformSolutions/generator-symphony/blob/master/generators/bdk/java/templates/spring/GifSlashHandler.java.ejs#L6
- https://github.com/SymphonyPlatformSolutions/generator-symphony/blob/master/generators/bdk/java/templates/spring/OnUserJoinedRoomListener.java.ejs#L3
- https://github.com/SymphonyPlatformSolutions/generator-symphony/blob/master/generators/bdk/java/templates/common/GifFormActivity.java.ejs#L8

Please create a ticket to update this repo accordingly.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -178,11 +181,22 @@ public V4Message send(@Nonnull V4Stream stream, @Nonnull String message) {
...
 
+  /**
+   * Sends a templated to the stream ID of the passed {@link V4Stream} object.
+   *
+   * @param stream     the stream to send the message to
+   * @param template   the template name to be used to produce the message
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @throws TemplateException
+   */
+  public V4Message sendWithTemplate(@Nonnull V4Stream stream, @Nonnull String template) {
```
symphony-thibault, line 196: Changing name of this method will also break `generator-symphony`.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -206,18 +231,46 @@ public V4Message send(@Nonnull String streamId, @Nonnull String template, Object
...
+   * Sends a message to the stream ID passed in parameter.
+   *
+   * @param streamId    the ID of the stream to send the message to
+   * @param message     the message payload in MessageML
+   * @param data        the data to be send with the message
+   * @param attachment  the attachment of the message
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @see <a href="https://developers.symphony.com/restapi/reference#create-message-v4">Create Message v4</a>
+   */
+  public V4Message send(@Nonnull String streamId, @Nonnull String message, String data, File attachment) {
```
symphony-thibault, line 247: I'd prefer to have an `InputStream attachment` or/and `byte[] attachment` instead of `File attachment`. 

`byte[]` is optional as developers can still easily `new ByteArrayInputStream(bytes)`

symphony-thibault: In method implementation, the ephemeral `File` object would be temporary created in `java.io.tmpdir` folder. Could you however take a look to the `org.openapi.generator` plugin to see if there is way for API clients to accept something else than a File object.

symphony-thibault (commented):
I do have some concerns regarding breaking changes you've introduced here. Even if we are still in BETA, could we try to make a bit more smoother through deprecation mechanism? 

We will clearly define the policy on next spring though (cf. `APP-3114`)

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/TemplateException.java
```diff
@@ -10,7 +10,7 @@
  * if template is malformed or if some parameters are missing when calling {@link Template#process(Object)}
  */
 @API(status = API.Status.STABLE)
-public class TemplateException extends Exception {
+public class TemplateException extends RuntimeException {
```
symphony-thibault, line 13: We might to update documentation, please take a look

symphony-hong:
@symphony-thibault yes clearly this PR will produce some breaking changes as we are still in BETA. I also supposed to open a ticket to update the generator after this PR is finished and merged.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: the problem here is that I don't think we can get the file type from an `InputStream` object. If we don't specify the extension file type in the file name, we will get the 403 Exception `File type not allowed`

symphony-hong: but how can we know by what we should replace`file-name.ext` since we get no information about the filename from the `InputStream`?

symphony-bdk-template/symphony-bdk-template-api/src/main/java/com/symphony/bdk/template/api/TemplateException.java
symphony-hong: I didn't find any existing documentation mentioning the `TemplateException`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-thibault: Then developer should provide it, or at least just the extension (name could be just `UUID.randomUUID()`), I don't know

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-youri: Do we plan to introduce a message object (to avoid passing many parameters and having many methods to send a message)? If so we will probably go back to a single send method no?

symphony-youri (commented):
_Make the TemplateException extends RuntimeException -> Remove all the throws TemplateException._
-> we see that from the change what would be interesting to state in the commit message is why :)? (the what is visible from the diff, the why is interesting to document in git logs so we can understand the decisions later when we look back at the code ;))

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: @symphony-youri yes I think it's a better solution but we can consider it next sprint?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-youri: Spring has the concept of MultipartFile that has a file name, a mime type and an input stream
Maybe we could have this sort of object to give us more flexibility (with constructors taking file, bytes,...)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-youri: Yes I'm just pointing that we might change the API here, update the doc and generator for it and then do it again next sprint so maybe it is easier to do it this sprint (how given how fast you are I'm pretty sure it could be done:))

symphony-thibault: We have a lot of things already plan for next sprint, could you tackle this idea as part of this PR? As it would also really reduce the complexity of this class.

symphony-hong:
@symphony-thibault @symphony-youri I created a Message model to be used in the message service, added two methods using this model as well.
I still keep the old methods and will refactor the messsage service after we agree that the Message model is okay to use.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/AttachmentType.java
```diff
@@ -0,0 +1,41 @@
+package com.symphony.bdk.core.service.message.model;
+
+import org.apiguardian.api.API;
+
+@API(status = API.Status.EXPERIMENTAL)
+public enum AttachmentType {
```
symphony-thibault, line 6: This list is not fixed : https://developers.symphony.com/restapi/reference#attachment-types :) Therefore I'm not sure that we can really provide an `enum` here

symphony-hong: I'm also wondering about this one. But I really don't like to make another API call to just create a message.

symphony-hong: or maybe we can let the user specify the whatever extension file type they want?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.symphony.bdk.template.api.Template;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import org.apiguardian.api.API;
+
+@Getter
+@Setter
```
symphony-youri, line 13: I would make the fields final to make this object immutable

symphony-youri: Mmm wait we also want to chain things, should we use a builder here?

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/ComplexMessageExampleMain.java
```diff
@@ -0,0 +1,57 @@
...
+                AttachmentType.DOC);
+    Template template = bdk.messages().templates().newTemplateFromClasspath("/complex-message.ftl");
+    Message message = Message.fromTemplate(template, Collections.singletonMap("name", "Freemarker")).attachment(attachment);
+    bdk.datafeed().subscribe(new RealTimeEventListener() {
+      @Override
+      public void onMessageSent(V4Initiator initiator, V4MessageSent event) {
+        try {
+          bdk.messages().send(event.getMessage().getStream(), message);
+        } catch (IOException e) {
+          e.printStackTrace();
```
symphony-youri, line 41: logger? (not that is very important for examples but as we are creating it lets use it)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Attachment.java
```diff
@@ -0,0 +1,18 @@
...
+import java.io.InputStream;
+
+@Getter
+@Setter
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class Attachment {
+
+  private InputStream inputStream;
+  private AttachmentType attachmentType;
```
symphony-youri, line 17: don't we need to distinguish the type and the file name?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -218,6 +248,92 @@ public V4Message send(@Nonnull String streamId, @Nonnull String message) {
...
+  /**
+   * Sends a message to the stream ID passed in parameter.
+   *
+   * @param stream    the stream to send the message to
+   * @param message   the message to send to the stream
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @throws IOException if the attachment file cannot be created.
+   * @see <a href="https://developers.symphony.com/restapi/reference#create-message-v4">Create Message v4</a>
+   */
+  public V4Message send(@Nonnull V4Stream stream, @Nonnull Message message) throws IOException {
```
symphony-youri, line 333: with this new method we can avoid adding the other ones sendWithTemplate and other with attachement, can't we?

symphony-hong: yes indeed

symphony-hong: we can even remove all the existing sendWithTemplate if you think it's redudant

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/AttachmentType.java
symphony-thibault: > I'm also wondering about this one. But I really don't like to make another API call to just create a message.

Of course we won't make this call for each message creation. However, I think it is also import to adopt a "fail-fast" strategy directly at BDK level, this would avoid overloading Agent/SBE with requests that will in the end fail because the file type is not allowed. 

I see 2 strategies that we could implement: 
- lazy load the list of allowed attachments during first call to send message
- load this list at BDK startup (both in `SymphonyBdk` and in Spring Boot starter)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
```diff
@@ -0,0 +1,37 @@
...
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class Message {
+
+  private String content;
+  private String data;
```
symphony-thibault, line 19: `data` could also be `JsonNode` or any `Object` that we will then serialize to a JSON string value

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import org.apiguardian.api.API;
+
+@Getter
+@Setter
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class Message {
```
symphony-thibault, line 16: We also need the optional `version` field, default value is `2.0`.

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/ComplexMessageExampleMain.java
```diff
@@ -0,0 +1,57 @@
...
+            .attachmentType(
+                AttachmentType.DOC);
+    Template template = bdk.messages().templates().newTemplateFromClasspath("/complex-message.ftl");
+    Message message = Message.fromTemplate(template, Collections.singletonMap("name", "Freemarker")).attachment(attachment);
+    bdk.datafeed().subscribe(new RealTimeEventListener() {
+      @Override
+      public void onMessageSent(V4Initiator initiator, V4MessageSent event) {
+        try {
+          bdk.messages().send(event.getMessage().getStream(), message);
+        } catch (IOException e) {
```
symphony-thibault, line 40: Do we really want to force developers to handle `IOException` each time they want to send a message? Instead, could we wrap it as a more functional `RuntimeException`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
symphony-hong: I don't really understand the purpose of `version` field here.

symphony-thibault: That's the version of the `MessageML` format

symphony-thibault: https://github.com/symphonyoss/messageml-utils/#usage

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Attachment.java
symphony-elias: Yes IMHO we should be able to let the user specify a file name.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -218,6 +248,56 @@ public V4Message send(@Nonnull String streamId, @Nonnull String message) {
...
+   * @param message     the message to send to the stream
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @see <a href="https://developers.symphony.com/restapi/reference#create-message-v4">Create Message v4</a>
+   */
+  public V4Message send(@Nonnull String streamId, @Nonnull Message message) {
+    File tempFile;
+    Attachment attachment = message.getAttachment();
+    if (attachment != null) {
+      try {
+        String[] fileNameArrays = attachment.fileName().split("\\.");
```
symphony-elias, line 264: If I understand well, the split is only there to validate the string format of `attachment.fileName()`. If so, you could extract this logic to a `validateAttachmentFileName` method (or with similar name) and reuse `attachment.fileName()` when you create a tempFile.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -218,6 +248,56 @@ public V4Message send(@Nonnull String streamId, @Nonnull String message) {
...
+        if (fileNameArrays.length < 2) {
+          throw new MessageCreationException("Invalid attachment's file name.");
+        }
+        tempFile = File.createTempFile(fileNameArrays[0], "." + fileNameArrays[fileNameArrays.length - 1]);
```
symphony-elias, line 268: I would put `fileNameArrays[1]` since you validated the length above. In my opinion putting index 1 is clearer.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -218,6 +248,56 @@ public V4Message send(@Nonnull String streamId, @Nonnull String message) {
...
+   * Sends a message to the stream ID passed in parameter.
+   *
+   * @param streamId    the ID of the stream to send the message to
+   * @param message     the message to send to the stream
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @see <a href="https://developers.symphony.com/restapi/reference#create-message-v4">Create Message v4</a>
+   */
+  public V4Message send(@Nonnull String streamId, @Nonnull Message message) {
+    File tempFile;
+    Attachment attachment = message.getAttachment();
```
symphony-elias, line 261: Please extract this logic or part of this logic to create a tempFile in a dedicated method

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: I am not sure if the user passes a name with multiple `.` like `file.name.ext`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-elias, line 264: Maybe you could put this validation logic and this parsing in the Attachment class

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
symphony-elias: Done in the MessageBuilder class.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: I actually don't like this logic. I would prefer user specify file name and extension in 2 different fields. WDYT?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/MessageBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public static MessageBuilder fromTemplate(Template template, Object parameters) {
+    return new MessageBuilder().content(template.process(parameters));
+  }
+
+  public static MessageBuilder fromTemplate(Template template) {
+    return new MessageBuilder().content(template.process(emptyMap()));
+  }
+
+  public MessageBuilder data(JsonNode data) {
```
symphony-elias, line 46: We can remove this one as the method below takes an `Object` as paremeter.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -386,4 +399,25 @@ private static Long getEpochMillis(Instant instant) {
   private <T> T executeAndRetry(String name, SupplierWithApiException<T> supplier) {
     return RetryWithRecovery.executeAndRetry(retryBuilder, name, supplier);
   }
+
+
+  private File createTemporaryAttachmentFile(Attachment attachment) {
+    File tempFile;
+    if (attachment != null) {
```
symphony-elias, line 406: Having an early return will slightly increase readability in my opinion, like:
`if (attachment == null) { return null; }` and then the regular case.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -165,59 +170,67 @@ public TemplateEngine templates() {
...
 
   /**
-   * Sends a templated to the stream ID of the passed {@link V4Stream} object.
+   * Sends a message to the stream ID passed in parameter.
    *
-   * @param stream     the stream to send the message to
-   * @param template   the template name to be used to produce the message
-   * @param parameters the parameters to pass to the template to produce the message to be sent
+   * @param streamId the ID of the stream to send the message to
+   * @param message  the message payload in MessageML
```
symphony-thibault, line 182: Use Javadoc [`@deprecated`](https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html#javadoc_tag) tag to mention that this method will be replaced by `send(String, Message)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -165,59 +170,67 @@ public TemplateEngine templates() {
    * @return a {@link V4Message} object containing the details of the sent message
    * @see <a href="https://developers.symphony.com/restapi/reference#create-message-v4">Create Message v4</a>
    */
+  @Deprecated
```
symphony-thibault, line 173: Use Javadoc [`@deprecated`](https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html#javadoc_tag) tag to mention that this method will be replaced by `send(V4Stream, Message)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Attachment.java
```diff
@@ -0,0 +1,18 @@
...
+import java.io.InputStream;
+
+@Getter
+@Setter
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class Attachment {
+
+  private InputStream inputStream;
+  private String fileName;
```
symphony-thibault, line 17: `filename`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/MessageBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+  public static MessageBuilder fromTemplate(Template template) {
+    return new MessageBuilder().content(template.process(emptyMap()));
+  }
+
+  /**
+   * Add data to the message.
+   * @param   data Serializable data object.
+   * @return  this builder with the data configured.
+   */
+  public MessageBuilder data(Object data) {
```
symphony-thibault, line 69: We also need `data(String data)` and `data(JsonNode json)`

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/ComplexMessageExampleMain.java
```diff
@@ -0,0 +1,53 @@
...
+
+    final SymphonyBdk bdk = new SymphonyBdk(loadFromSymphonyDir("config.yaml"));
+
+    Attachment attachment =
+        new Attachment().inputStream(IOUtils.toInputStream("This is string", StandardCharsets.UTF_8))
+            .fileName("test.doc");
+    Template template = bdk.messages().templates().newTemplateFromClasspath("/complex-message.ftl");
+    Message message = MessageBuilder.fromTemplate(template, Collections.singletonMap("name", "Freemarker"))
+        .attachment(attachment)
+        .build();
```
symphony-thibault, line 34: That's not very clear, for `Message` we have a `MessageBuilder`, but before we have to `new Attachment()`

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/ComplexMessageExampleMain.java
```diff
@@ -0,0 +1,53 @@
...
+    bdk.datafeed().subscribe(new RealTimeEventListener() {
+      @Override
+      public void onMessageSent(V4Initiator initiator, V4MessageSent event) {
+        bdk.messages().send(event.getMessage().getStream(), message);
```
symphony-thibault, line 38: That would be nice to show that the `complex-message.ftl` contains event information

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/MessageBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+@Getter
+@Setter
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class MessageBuilder {
+
+  private static final ObjectMapper MAPPER = new JsonMapper();
+
+  private String version = "2.0";
```
symphony-thibault, line 28: Do we have a unit test that ensures that default version is `2.0`?

symphony-thibault: Add Javadoc comment explaining what is this field

symphony-thibault:
I'd ❤️ top have: 
```java
final InputStream toto = new FileInputStream("foobar.txt");
final V4Stream stream = event.getStream();

final V4Message messageSent = bdk.message().with()
    .template("/templates/welcome.ftl", values)
    .attachment(foobar, "toto.txt")
    .data(singletonMap("key", "value"))
    .send(stream);
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-youri: Should not we remove it at once since we are in beta?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/exception/MessageCreationException.java
```diff
@@ -0,0 +1,18 @@
+package com.symphony.bdk.core.service.message.exception;
+
+import org.apiguardian.api.API;
+
+/**
+ * Exception thrown when a {@link com.symphony.bdk.core.service.message.model.Message} is failed to create.
+ */
+@API(status = API.Status.EXPERIMENTAL)
```
symphony-thibault, line 8: This has to be `STABLE`

symphony-hong:
@symphony-thibault is it exactly what you suggested? https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/pull/277/commits/865bf89b2318da94ad8bde2bbf75ffde22ac6ed2#diff-1f276cf993a147eb9c7b918053efb18dL38

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -164,60 +174,70 @@ public TemplateEngine templates() {
...
         streamId,
         authSession.getSessionToken(),
         authSession.getKeyManagerToken(),
-        message,
-        null,
-        null,
+        message.getContent(),
+        message.getData(),
         null,
+        tempFile,
```
symphony-thibault, line 224: `this.createTemporaryAttachmentFile(message.getAttachment())` directly here

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Attachment.java
```diff
@@ -0,0 +1,18 @@
+package com.symphony.bdk.core.service.message.model;
+
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+import org.apiguardian.api.API;
+
+import java.io.InputStream;
+
+@Getter
```
symphony-thibault, line 10: javadoc

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/MessageBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+    this.messageService = messageService;
+  }
+
+  /**
+   * Add messageML content to the message.
+   *
+   * @param   message    messageML.
+   * @return  this builder with the content configured.
+   */
+  public MessageBuilder messageML(String message) {
```
symphony-thibault, line 50: why not just `content`? As you mention in the Javadoc that it is expecting MessageML

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/MessageBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+  public MessageBuilder data(JsonNode data) {
+    try {
+      this.data = MAPPER.writeValueAsString(data);
+      return this;
+    } catch (JsonProcessingException e) {
+      throw new MessageCreationException("Failed to parse data to Json string", e);
+    }
+  }
+
+  public MessageBuilder attachment(InputStream inputStream, String filename) {
```
symphony-thibault, line 106: javadoc

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -93,6 +99,10 @@ public TemplateEngine templates() {
     return templateEngine;
   }
 
+  public MessageBuilder builder() {
```
symphony-thibault, line 102: javadoc

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Attachment.java
symphony-youri: Can we keep this class immutable too?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
```diff
@@ -0,0 +1,30 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apiguardian.api.API;
+
+/**
+ * Message model to be used in {@link com.symphony.bdk.core.service.message.MessageService#send(V4Stream, Message)}
+ */
+@Getter
+@Setter
```
symphony-youri, line 13: Can we keep it immutable?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
```diff
@@ -0,0 +1,30 @@
...
+@API(status = API.Status.EXPERIMENTAL)
+public class Message {
+
+  // The version of the MessageML format
+  private String version;
```
symphony-youri, line 18: when is the version used?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: I don't know why yet but if I use `this.createTemporaryAttachmentFile(message.getAttachment())`, method will throw a `RuntimeException` instead of `MessageCreationException` and the unittest will fail

symphony-hong: because the retry throws every `Throwable` as `RuntimeException`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -386,4 +406,22 @@ private static Long getEpochMillis(Instant instant) {
...
+    if (attachment == null) {
+      return null;
+    }
+    try {
+      String[] fileNameArrays = attachment.filename().split("\\.");
+      if (fileNameArrays.length < 2) {
+        throw new MessageCreationException("Invalid attachment's file name.");
+      }
+      File tempFile = File.createTempFile(fileNameArrays[0], "." + fileNameArrays[fileNameArrays.length - 1]);
+      tempFile.deleteOnExit();
```
symphony-youri, line 420: the file will be deleted only after the bot exits which could be never (if the bot is robust enough ::) 
If seems it could a risk in term of security, for instance if the bot is sending sensitive files, they would stay on disk for a long time in a temp folder that could be accessible to other users
I think we should find a way to not use a file at all (or maybe it is the only way to make it work for large files but in that case we should perhaps make it explicit that a folder is used for the files, making it appear in config for instance that the user is aware of the risk)
we could also try to delete the file after the call is made

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -386,4 +406,22 @@ private static Long getEpochMillis(Instant instant) {
...
+  private File createTemporaryAttachmentFile(Attachment attachment) {
+    if (attachment == null) {
+      return null;
+    }
+    try {
+      String[] fileNameArrays = attachment.filename().split("\\.");
+      if (fileNameArrays.length < 2) {
+        throw new MessageCreationException("Invalid attachment's file name.");
+      }
+      File tempFile = File.createTempFile(fileNameArrays[0], "." + fileNameArrays[fileNameArrays.length - 1]);
```
symphony-youri, line 419: aren't we creating a file using the filename that is already set in the name? (well dropping part of it actually if it has multiple dots)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -386,4 +406,22 @@ private static Long getEpochMillis(Instant instant) {
...
     return RetryWithRecovery.executeAndRetry(retryBuilder, name, supplier);
   }
+  
+  private File createTemporaryAttachmentFile(Attachment attachment) {
+    if (attachment == null) {
+      return null;
+    }
+    try {
+      String[] fileNameArrays = attachment.filename().split("\\.");
+      if (fileNameArrays.length < 2) {
```
symphony-youri, line 416: isn't it something we could check when building the attachment as to create only valid attachment?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: we cannot create a temp file with a specific name. The temp file created by the `File#createTempFile` always has a random string in its name

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: And I would create the temp file outside the retry to avoid new file created if the retry recovery is triggered

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-elias: +1 for deleting the file after the call is made

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
symphony-elias: This is an optional field in the endpoint : https://developers.symphony.com/restapi/reference#create-message-v4

symphony-thibault (approved):
Good job 👍 Will be completed on next sprint with the ability to send multiple attachments, I'll create another ticket

PR #285, opened 2020-10-13
APP-3108: Implement Circle of Trust controller in app-starter

symphony-hong:
### Ticket
[APP-3108](https://perzoinc.atlassian.net/browse/APP-3108)

### Description
Move the method AppAuthSession#validateJwt to ExtensionAppAuthenticator#validateJwt to have the ExtensionAppAuthenticator performing all steps of the Circle Of Trust.

Change UserClaim.id to Long type because the Symphony UserId is always a Long number.

Extension App Spring Boot Starter is developed under the Core Starter, the auto-configuration is reuse the SymphonyBdkCoreProperties with auto-configured bean define in BdkExtAppConfig and BdkExtAppControllerConfig
BdkExtAppConfig is used for configuring the beans of ExtensionAppAuthenticator
BdkExtAppControllerConfig is used for configuring the RestController AppAuthController, enabled by the configuration: bdk.app.auth.enabled=true|false

The App Spring Boot Starter example creates a Spring Boot Application using the App Spring Boot Starter

**Unittest and documentation are not included yet.**

### Dependencies

### Checklist


symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/controller/AppAuthController.java
```diff
@@ -0,0 +1,112 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@Slf4j
+@RestController
+@RequestMapping(value = {"/application", "/applications"})
```
symphony-thibault, line 29: I believe that it will be import for developers to quickly identify which endpoints are provided by this starter or not. So we should make it part of the request mapping: `@RequestMapping("/bdk/v1/app/auth")` (as well as versioning)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/controller/AppAuthController.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @PostMapping("tokens/validate")
+  public ResponseEntity<?> validateTokens(@RequestBody AppToken appToken) {
+    log.debug("App auth step 2: Validating tokens");
+    if (extensionAppAuthenticator.validateTokens(appToken.getAppToken(), appToken.getSymphonyToken())) {
+      return ResponseEntity.ok().build();
+    }
+
+    return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
```
symphony-thibault, line 79: We should provide a `@ControllerAdvice` and throw exceptions, instead of returning `ResponseEntity`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/controller/AppAuthController.java
```diff
@@ -0,0 +1,112 @@
...
+      log.info("Authentication process failed");
+      return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
+    }
+  }
+
+  @PostMapping("tokens/validate")
+  public ResponseEntity<?> validateTokens(@RequestBody AppToken appToken) {
+    log.debug("App auth step 2: Validating tokens");
+    if (extensionAppAuthenticator.validateTokens(appToken.getAppToken(), appToken.getSymphonyToken())) {
+      return ResponseEntity.ok().build();
```
symphony-thibault, line 76: no content (204) instead of ok (200)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/controller/AppAuthController.java
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      AppAuthSession authSession = extensionAppAuthenticator.authenticateExtensionApp(appInfo.getAppToken());
+      return ResponseEntity.ok(authSession);
+    } catch (AuthUnauthorizedException authUnauthorizedException) {
+      log.info("Authentication process failed");
+      return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
+    }
+  }
+
+  @PostMapping("tokens/validate")
```
symphony-thibault, line 72: just `@PostMapping("/tokens")`

symphony-bdk-examples/bdk-app-spring-boot-example/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    compileOnly 'org.projectlombok:lombok'
+    annotationProcessor 'org.projectlombok:lombok'
+
+    testImplementation 'com.tngtech.archunit:archunit-junit5'
+    testImplementation('org.springframework.boot:spring-boot-starter-test') {
+        exclude group: 'org.junit.vintage', module: 'junit-vintage-engine'
+    }
+}
+
+uploadArchives.enabled = false
```
symphony-thibault, line 23: `jacocoTestCoverageVerification` plugin is not configured

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/java/com/symphony/bdk/examples/app/spring/AppSpringApplication.java
```diff
@@ -0,0 +1,13 @@
+package com.symphony.bdk.examples.app.spring;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class AppSpringApplication {
```
symphony-thibault, line 7: I'd like to see a real application in action in this module, it could just be a simple `index.html` + `main.js` (under `src/main/resources/static`) in pure Javascript (@symphony-adnane could help 😃) that calls the `AppAuthController` endpoints

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/controller/AppAuthController.java
```diff
@@ -0,0 +1,112 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@Slf4j
+@RestController
+@RequestMapping(value = {"/application", "/applications"})
+@ConditionalOnProperty(name = "bdk.app.auth.enabled", havingValue = "true")
+public class AppAuthController {
+
+  @Value("${jwt-cookie.enable:false}")
```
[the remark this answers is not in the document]
symphony-hong: I didn't get your point here. I thought the variable with `@Value` will automatically get value from `application.yaml`?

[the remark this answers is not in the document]
symphony-thibault: All available properties should be part of a `@ConfigurationProperties(prefix = "bdk-app")` class (cf. [SymphonyBdkCoreProperties.java](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/SymphonyBdkCoreProperties.java))

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/java/com/symphony/bdk/examples/app/spring/AppSpringApplication.java
symphony-hong: I would prefer to do it in the next PR 😄

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/controller/AppAuthController.java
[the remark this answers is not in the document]
symphony-hong: So I guess `bdk.app.auth.jwt-cookie.enabled` and `bdk.app.auth.jwt-cookie.expire-in` is okay, right? (in my last commit)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+    if (jwtCookieEnable) {
+      response.addCookie(jwtCookie(jwt, request.getContextPath()));
+    }
+    return ResponseEntity.ok(userId);
+  }
+
+  private Cookie jwtCookie(String jwt, String path) {
+    Cookie jwtCookie = new Cookie("userJwt", jwt);
+
+    // 1 day
```
symphony-youri, line 78: comment to remove?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+  public ResponseEntity<Long> validateJwt(@Valid @RequestBody JwtInfo jwtInfo, HttpServletRequest request,
+      HttpServletResponse response) throws AuthInitializationException {
+    log.debug("App auth step 3: Validate jwt");
+    String jwt = jwtInfo.getJwt();
+    UserClaim userClaim = extensionAppAuthenticator.validateJwt(jwtInfo.getJwt());
+    Long userId = userClaim.getId();
+    if (jwtCookieEnable) {
+      response.addCookie(jwtCookie(jwt, request.getContextPath()));
+    }
+    return ResponseEntity.ok(userId);
```
symphony-youri, line 72: Can we return a JSON body instead (more flexible in the future) (if we are allowed to change the format)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+  public AppAuthController(ExtensionAppAuthenticator extensionAppAuthenticator) {
+    this.extensionAppAuthenticator = extensionAppAuthenticator;
+  }
+
+  @PostMapping
+  public ResponseEntity<AppAuthSession> authenticate(@Valid @RequestBody AppInfo appInfo)
+      throws AuthUnauthorizedException {
+    log.debug("App auth step 1: Initializing extension app authentication");
+
+    AppAuthSession authSession = extensionAppAuthenticator.authenticateExtensionApp(appInfo.getAppToken());
```
symphony-youri, line 48: AppAuthSession is part of our model, here I would perhaps introduce a DTO for the reponse to decouple both (to avoid returning new information if we evolve AppAuthSession for instance)

symphony-youri: For instance here we are returning the symphony token which is probably not what we want at all right?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.validation.Valid;
+
+@Slf4j
+@RestController
+@RequestMapping("/bdk/v1/app/auth")
```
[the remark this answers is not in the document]
symphony-youri: If we can change the URLs then I would wonder if we should go the REST way trying to have resources
or we keep it more in a RPC style with verbs
here we are mixing both I would say with /auth /tokens for instance

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/java/com/symphony/bdk/examples/app/spring/AppSpringApplication.java
symphony-youri: +1 that's a must have as we are trying to replace https://github.com/symphonyoss/app-auth-example
(and of course yes it can be done after this PR)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+  private Integer jwtCookieExpireInSec;
+
+  private final ExtensionAppAuthenticator extensionAppAuthenticator;
+
+  public AppAuthController(ExtensionAppAuthenticator extensionAppAuthenticator) {
+    this.extensionAppAuthenticator = extensionAppAuthenticator;
+  }
+
+  @PostMapping
+  public ResponseEntity<AppAuthSession> authenticate(@Valid @RequestBody AppInfo appInfo)
```
symphony-youri, line 44: the controller can be tested with spring mock mvc (but maybe you already that coming next :) )

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+    log.debug("App auth step 1: Initializing extension app authentication");
+
+    AppAuthSession authSession = extensionAppAuthenticator.authenticateExtensionApp(appInfo.getAppToken());
+    return ResponseEntity.ok(authSession);
+  }
+
+  @PostMapping("/tokens")
+  public ResponseEntity<?> validateTokens(@Valid @RequestBody AppToken appToken) {
+    log.debug("App auth step 2: Validating tokens");
+    if (extensionAppAuthenticator.validateTokens(appToken.getAppToken(), appToken.getSymphonyToken())) {
```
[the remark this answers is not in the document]
symphony-thibault: I have no idea of why it is like that. Regarding the [Circle Of Trust](https://developers.symphony.com/extension/docs/application-authentication#section-application-authentication-sequence), I'd say that it's because storing those tokens is App Backend responsibility, by definition...

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/AppToken.java
```diff
@@ -0,0 +1,16 @@
...
+
+import javax.validation.constraints.NotBlank;
+
+@Data
+@NoArgsConstructor
+public class AppToken {
+
+  @NotBlank(message = "App Token is mandatory")
+  private String appToken;
+  @NotBlank(message = "Symphony is mandatory")
```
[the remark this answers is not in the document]
symphony-hong: Symphony token is the token returned by SBE after we authenticate to a extension app.
https://developers.symphony.com/extension/docs/application-authentication#section-initiate-backend-authentication

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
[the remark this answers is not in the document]
symphony-hong: so it means `/bdk/v1/app/auth` and `/bdk/v1/app/tokens`?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
symphony-hong: the symphony token should be returned because it's needed for the second step. I guess we can use the AppToken (which is used for the Request body in step 2) for the response?

symphony-youri: In the legacy one it is not returned

Here it does not show either https://developers.symphony.com/extension/docs/application-authentication#section-application-authentication-sequence I think the idea of get the Ts (symphony token) from the frontend directly and then check it matches with the one stored by the app

Otherwise we by pass the part of the circle happing before step 17

@symphony-thibault do you agree?

symphony-thibault: Correct, `Ts` at [step 17](https://developers.symphony.com/extension/docs/application-authentication#section-application-authentication-sequence) is retrieved from the Symphony Frontend, not returned step 11 by App Backend

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/java/com/symphony/bdk/examples/app/spring/AppSpringApplication.java
symphony-thibault: You can do it as part of a separate PR once this one merged, but on the same ticket

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
```diff
@@ -0,0 +1,18 @@
...
+    username: fxbot-hong
+    privateKeyPath: ${user.home}/.symphony/keys/rsa-private-bot.pem
+
+  app:
+    auth.enabled: true
+    auth.jwt-cookie:
+      enabled: true
+      expire-in: 43200
+    appId: testapp-hongle
+    privateKeyPath: ${user.home}/.symphony/keys/app_privatekey.pem
```
symphony-thibault, line 14:
```yaml
app:
  appId: testapp-hongle
  privateKeyPath: ${user.home}/.symphony/keys/app_privatekey.pem
  auth:
    enabled: true # activates AppAuthController 
    jwt-cookie:
      enabled: true
      max-age: 43200 # in seconds
```

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
symphony-thibault, line 44: You can directly return `AppAuthSession`, `ResponseEntity<>` is not needed anymore as you now handle errors in `@ControllerAdvice`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
```diff
@@ -0,0 +1,87 @@
...
+
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.validation.Valid;
+
+@Slf4j
+@RestController
+@RequestMapping("/bdk/v1/app/auth")
+public class AppAuthController {
```
symphony-thibault, line 29: Why not `CircleOfTrustController` ?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+import java.time.LocalDateTime;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@ControllerAdvice
```
symphony-thibault, line 23: Is this annotation still required as you now inject it from `BdkExtAppControllerConfig`?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,59 @@
...
+public class GlobalControllerExceptionHandler extends ResponseEntityExceptionHandler {
+
+  public GlobalControllerExceptionHandler() {
+    super();
+  }
+
+  @ExceptionHandler({
+      AuthInitializationException.class,
+      AuthUnauthorizedException.class
```
symphony-thibault, line 32: Those exceptions should be handled and wrapped in a service layer. Then this layer would throw an `AppAuthException extends RuntimeException` specific to this `app-starter`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,59 @@
...
+  })
+  public ResponseEntity<Object> handleUnauthorizedException(Exception e, WebRequest request) {
```
symphony-thibault, line 34: We need a dedicated error class: something like `BdkAppError` that would contain an `enum BdkAppErrorCode` describing all error codes potentially returned by the starter.

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
symphony-hong: I would use `jwtCookie` and `maxAge` instead to make the format consistent

symphony-thibault: You could check, but if I remember well both are supported by Spring yaml parser

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/GlobalControllerExceptionHandler.java
symphony-hong: It will not work if the annotation is removed

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/AppAuthController.java
symphony-hong: I am actually struggling using spring mock mvc to test a RestController in a starter. Can you give me an small example how it works?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -0,0 +1,103 @@
...
+@Slf4j
+@RestController
+@RequestMapping("/bdk/v1/app")
+public class CircleOfTrustController {
+
+  @Value("${bdk.app.auth.jwtCookie.enabled:false}")
+  private Boolean jwtCookieEnable;
+
+  @Value("${bdk.app.auth.jwtCookie.maxAge:86400}")
+  private Integer jwtCookieMaxAgeInSec;
```
symphony-thibault, line 39: This is a bad practice, this starter should provide its own properties, like [SymphonyBdkCoreProperties](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/SymphonyBdkCoreProperties.java). 

So that you could directly inject a `SymphonyBdkAppProperties` in your class.

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -0,0 +1,101 @@
...
+  public CircleOfTrustController(SymphonyBdkAppProperties properties, ExtensionAppAuthenticator extensionAppAuthenticator) {
+    this.properties = properties;
+    this.extensionAppAuthenticator = extensionAppAuthenticator;
+  }
+
+  @PostMapping("/auth")
+  public AppToken authenticate(@Valid @RequestBody AppInfo appInfo) {
+    log.debug("App auth step 1: Initializing extension app authentication");
+
+    log.debug(properties.getAuth().getJwtCookie().getMaxAge().toString());
```
symphony-elias, line 47: Do we want to keep this line ?

symphony-hong: Oops! 😬 I forgot to remove it

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/SymphonyBdkAppProperties.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Configuration Properties for the Symphony BDK Extension App.
+ */
+@Getter
+@Setter
+@ConfigurationProperties(prefix = "bdk.app")
+public class SymphonyBdkAppProperties extends BdkExtAppConfig {
+  private AppAuthProperties auth;
+}
```
symphony-thibault, line 18: Not sure that this is the right way to manage starters properties. IMHO we should have 2 separate entries in `application.yaml` file: 
```yaml
bdk:     # core starter config
   host: acme.symphony.com
bdk-app: # ext app starter config
   auth: 
     enabled: true
     jwtCookie: 
        enabled: true
        expireIn: 43200
```

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -0,0 +1,112 @@
...
+  private final ExtensionAppAuthenticator extensionAppAuthenticator;
+
+  public CircleOfTrustController(SymphonyBdkAppProperties properties, ExtensionAppAuthenticator extensionAppAuthenticator) {
+    this.properties = properties;
+    this.extensionAppAuthenticator = extensionAppAuthenticator;
+  }
+
+  @PostMapping("/auth")
+  public AppToken authenticate() {
+    log.debug("App auth step 1: Initializing extension app authentication");
```
symphony-thibault, line 52: we need better logs, revamp messages

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -0,0 +1,112 @@
...
+      return appToken;
+    } catch (AuthUnauthorizedException e) {
+      throw new AppAuthException("Failed to authenticate the extension app", e);
+    }
+  }
+
+  @PostMapping("/tokens")
+  public ResponseEntity<?> validateTokens(@Valid @RequestBody TokenPair appToken) {
+    log.debug("App auth step 2: Validating tokens");
+    if (extensionAppAuthenticator.validateTokens(appToken.getAppToken(), appToken.getSymphonyToken())) {
```
symphony-thibault, line 69: if not, throw exception

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -0,0 +1,112 @@
...
+      return ResponseEntity.noContent().build();
+    }
+
+    throw new InvalidTokenException("Failed to validate the app token");
```
symphony-thibault, line 73: return `void` and annotate method with `@ResponseStatus(HttpStatus.NO_CONTENT)`. BWT, do we really want to return nothing? By experience I know that FE developers usually complain about that

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      String token = this.generateAppToken();
+      AppAuthSession authSession = extensionAppAuthenticator.authenticateExtensionApp(token);
+      AppToken appToken = new AppToken();
+      appToken.setAppToken(authSession.getAppToken());
+
+      return appToken;
+    } catch (AuthUnauthorizedException e) {
+      throw new AppAuthException("Failed to authenticate the extension app", e);
+    }
```
symphony-thibault, line 63: That'd be nice to provide a service layer. Usually, a controller method should perform a single call to a service method

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/AppToken.java
```diff
@@ -0,0 +1,9 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+import lombok.Data;
+
+@Data
+public class AppToken {
```
symphony-thibault, line 6: javadoc missing, document fields as well

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/BdkAppError.java
```diff
@@ -0,0 +1,15 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class BdkAppError {
```
symphony-thibault, line 10: javadoc missing, document fields as well

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/BdkAppErrorCode.java
```diff
@@ -0,0 +1,8 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+public enum BdkAppErrorCode {
```
symphony-thibault, line 3: javadoc missing

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/BdkAppErrorCode.java
```diff
@@ -0,0 +1,8 @@
...
+  UNAUTHORIZED,
```
symphony-thibault, line 4: I'd expect the enum value + an error message template:  
```java 
enum BdkAppErrorCode {
   UNAUTHORIZED("App {{appId}} is unauthorized to ...") // for example
}
```

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/JwtInfo.java
```diff
@@ -0,0 +1,14 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotBlank;
+
+@Data
+@NoArgsConstructor
+public class JwtInfo {
```
symphony-thibault, line 10: javadoc...

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/TokenPair.java
```diff
@@ -0,0 +1,16 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotBlank;
+
+@Data
+@NoArgsConstructor
+public class TokenPair {
```
symphony-thibault, line 10: javadoc...

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/UserId.java
```diff
@@ -0,0 +1,9 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+import lombok.Data;
+
+@Data
+public class UserId {
```
symphony-thibault, line 6: javadoc....

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/config/BdkExtAppControllerConfig.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.context.annotation.Bean;
+
+/**
+ * Configuration and injection of the main rest controllers for extension app APIs as beans within the Spring application context.
+ */
+public class BdkExtAppControllerConfig {
+
+  @Bean
+  @ConditionalOnProperty(name = "bdk.app.auth.enabled", havingValue = "true")
+  public CircleOfTrustController circleOfTrustController(SymphonyBdkAppProperties properties,
```
symphony-thibault, line 19: Since `CircleOfTrustController` is annotated by `@RestController`, are we sure that it's not injected twice in the spring application context?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/AppAuthException.java
```diff
@@ -0,0 +1,15 @@
+package com.symphony.bdk.app.spring.exception;
```
symphony-thibault, line 1: Should go to `auth` package right?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Global error handler for the runtime exception thrown by the RestController.
+ */
+@ControllerAdvice
+public class GlobalControllerExceptionHandler extends ResponseEntityExceptionHandler {
+
+  @ExceptionHandler({AppAuthException.class})
```
symphony-thibault, line 26: `@ExceptionHandler(AppAuthException.class)`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/InvalidJwtException.java
```diff
@@ -0,0 +1,15 @@
+package com.symphony.bdk.app.spring.exception;
+
+/**
+ * Thrown when the Jwt is not valid.
+ */
+public class InvalidJwtException extends RuntimeException {
```
symphony-thibault, line 6: You could provide a "more" generic exception API, something like: 
```java
public class BdkAppException extends RuntimeException {
   private final BdkErrorCodeEnum error;
}
```
That would also simplify the controller advice

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/JwtCookieProperties.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Configuration Properties for storing Jwt in cookie for the Extension App
+ */
+@Getter
+@Setter
+public class JwtCookieProperties {
+
+  private Boolean enabled = false;
+  private Integer maxAge = 24*60*60;
```
symphony-thibault, line 14: Declare a static `DEFAULT_MAX_AGE`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/SymphonyBdkAppAutoConfigurationTest.java
```diff
@@ -0,0 +1,46 @@
...
+            "bdk.app.auth.jwtCookies.maxAge=86400"
+        )
+        .withUserConfiguration(SymphonyBdkMockedConfiguration.class)
+        .withConfiguration(AutoConfigurations.of(SymphonyBdkAppAutoConfiguration.class));
+
+    contextRunner.run(context -> {
+
+      // verify that main beans have been injected
+      assertThat(context).hasSingleBean(SymphonyBdkAppAutoConfiguration.class);
+      assertThat(context).hasSingleBean(BdkExtAppControllerConfig.class);
```
symphony-thibault, line 43: Also check that `assertThat(context).hasSingleBean(CircleOfTrustController.class)`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/SymphonyBdkAppAutoConfigurationTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+            "bdk.keyManager.scheme=http",
+            "bdk.keyManager.host=localhost",
+
+            "bdk.bot.username=tibot",
+            "bdk.bot.privateKeyPath=classpath:/privatekey.pem",
+
+            "bdk.app.appId=test-app",
+            "bdk.app.privateKeyPath=classpath:/privatekey.pem",
+            "bdk.app.auth.enabled=true",
```
symphony-thibault, line 32: Add another test with `"bdk.app.auth.enabled=false"` and verify that `CircleOfTrustController` is not injected

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/auth/CircleOfTrustControllerTest.java
```diff
@@ -0,0 +1,176 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.MockitoAnnotations;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+
+@RunWith(SpringRunner.class)
```
symphony-thibault, line 36: This is for JUnit4, we use JUnit5 so it should be `@ExtendWith(SpringExtension.class)`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/auth/CircleOfTrustControllerTest.java
```diff
@@ -0,0 +1,176 @@
...
+public class CircleOfTrustControllerTest {
+
+  private MockMvc mockMvc;
+  private static final ObjectMapper MAPPER = new JsonMapper();
+
+  @MockBean SymphonyBdkAppProperties properties;
+  @MockBean ExtensionAppAuthenticator authenticator;
```
symphony-thibault, line 43: Shouldn't be mocked. Take a look [here](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/test/java/com/symphony/bdk/spring/SymphonyBdkMockedConfiguration.java)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
symphony-hong: I believed this is the "if not" case 🤔 
https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/pull/285/files/1e3ee5e59a3640b2c7e301351f33c827a062b98e#diff-14dbe7e9144179fc9fcb029156ff0bc09c09c1dd32770e449424bc561a8ed327R73

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
symphony-hong: Because this step is only validating if the tokens pair that user provides is correct, I don't know what should be returned here. (Maybe a message saying that the tokens are correct, but I think it's not very useful)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
symphony-hong: I saw the comment below! Nevermind 😄

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/auth/CircleOfTrustControllerTest.java
symphony-hong: I think we only want to test the logic within the Controller, so IMHO, mocking the authenticator is enough?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/auth/CircleOfTrustControllerTest.java
symphony-thibault: BWT, we should also exclude JUnit4 from classpath

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/config/BdkExtAppControllerConfig.java
symphony-hong: when I set `bdk.app.auth.enabled=false`, the Api returns 404 Not Found, so I guess the controller is not injected twice

docs/spring-boot/app-starter.md
```diff
@@ -0,0 +1,145 @@
...
+```groovy
+plugins {
+    id 'java-library'
+    id 'org.springframework.boot' version '2.3.4.RELEASE'
+}
+
+dependencies {
+    implementation platform('com.symphony.platformsolutions:symphony-bdk-bom:1.3.2.BETA')
+    
+    implementation 'com.symphony.platformsolutions:symphony-bdk--spring-boot-starter'
```
symphony-youri, line 74: extra --?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/resources/application.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+  bot:
+    username: tibot
+    privateKeyPath: classpath:/privatekey.pem
+
+  app:
+    auth.enabled: true
+    auth.jwtCookie:
+      enabled: true
+      expireIn: 43200
```
symphony-thibault, line 22: This `application.yaml` is not up-to-date

docs/spring-boot/app-starter.md
```diff
@@ -0,0 +1,145 @@
...
+The Symphony BDK for Java provides a _Starter_ module that aims to ease extension app backend developments within a
+[Spring Boot](https://spring.io/projects/spring-boot) application.
+ 
+## Features
+- Configure extension app through `application.yaml`
+- Expose a REST API to perform [Circle of Trust authentication](https://developers.symphony.com/extension/docs/application-authentication).
+
+## Installation
+
+The following listing shows the `pom.xml` file that has to be created when using Maven:
```
symphony-youri, line 11: isn't this part going to be redundant with the generator?

docs/spring-boot/app-starter.md
```diff
@@ -0,0 +1,145 @@
...
+
+    public static void main(String[] args) {
+        SpringApplication.run(ExtAppSpringApplication.class, args);
+    }
+}
+```
+
+## Circle of Trust
+
+By configuring the property `bdk.app.auth.enabled=true`, the Application backend will provide Apis for performing
```
symphony-youri, line 116: I would write it as APIs

docs/spring-boot/app-starter.md
```diff
@@ -0,0 +1,145 @@
...
+
+    Parameters:
+    ```json
+    {
+      "appToken": "your-app-token",
+      "symphonyToken": "symphony-token" // Symphony token returned from Symphony App Frontend
+    }
+    ``` 
+  
+- Validate jwt: `POST /bdk/v1/app/tokens`
```
symphony-youri, line 138: jwt -> JWT
Is the URL correct? (it is the same as above)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/ExtensionAppAuthenticator.java
```diff
@@ -39,4 +41,15 @@
    * @return true if symphonyToken associated to a previous session initiated with appToken
    */
   boolean validateTokens(String appToken, String symphonyToken);
+
+
+  /**
+   * Validates a jwt against the pod certificate
```
symphony-youri, line 47: apart from the code as JWT is an acronym I would stick with uppercase

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
```diff
@@ -0,0 +1,21 @@
+
+bdk:
+  host: devx1.symphony.com
```
symphony-youri, line 3: Maybe try not to put internal server address here (could be localhost or a more generic/public pod maybe?)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/UserId.java
symphony-youri: Can we use @Value to make it immutable?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/TokenPair.java
```diff
@@ -0,0 +1,26 @@
...
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotBlank;
+
+/**
+ * A pair of tokens used during the Circle of Trust authentication.
+ */
+@Data
```
symphony-youri, line 11: Same here as for UserId can we make it immutable?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/JwtCookieProperties.java
```diff
@@ -0,0 +1,17 @@
...
+ * Configuration Properties for storing Jwt in cookie for the Extension App
+ */
+@Getter
+@Setter
+public class JwtCookieProperties {
+
+  private static final Integer DEFAULT_MAX_AGE = 24*60*60;
+
+  private Boolean enabled = false;
+  private Integer maxAge = DEFAULT_MAX_AGE;
```
symphony-youri, line 16: We could try to use duration directly https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/service/JwtService.java
```diff
@@ -0,0 +1,32 @@
...
+import com.symphony.bdk.app.spring.auth.model.UserId;
+import com.symphony.bdk.app.spring.auth.model.exception.AppAuthException;
+import com.symphony.bdk.core.auth.ExtensionAppAuthenticator;
+import com.symphony.bdk.core.auth.exception.AuthInitializationException;
+import com.symphony.bdk.core.auth.jwt.UserClaim;
+
+import org.springframework.stereotype.Service;
+
+@Service
+public class JwtService {
```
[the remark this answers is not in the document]
symphony-youri: and perhaps move       AppAuthSession authSession = extensionAppAuthenticator.authenticateExtensionApp(token);
 to this CircleOfTrustService so that the controller uses this service only

docs/spring-boot/app-starter.md
```diff
@@ -0,0 +1,145 @@
...
+## Create a Simple Backend for Extension Application
+As a first step, you have to initialize your environment through the Spring Boot `src/main/resources/application.yaml` file: 
+```yaml
+bdk:
+    host: acme.symphony.com
+    bot:
+      username: bot-username
+      privateKeyPath: /path/to/rsa/privatekey.pem
+      
+    app:
```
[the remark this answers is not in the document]
symphony-hong: I'm not sure where we want to keep the appId and the privatekey, because the `CoreProperties` extending the BdkConfig so I guess the appId and privatekey should be under the bdk.

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/exception/AppAuthException.java
```diff
@@ -0,0 +1,25 @@
+package com.symphony.bdk.app.spring.auth.model.exception;
+
+import com.symphony.bdk.app.spring.auth.model.BdkAppErrorCode;
+
+/**
+ * Thrown when the extension app authentication failed.
+ */
+public class AppAuthException extends RuntimeException {
```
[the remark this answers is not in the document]
symphony-hong: So should we move it back outside the `auth` package since it is generic?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/TokenPair.java
symphony-hong: if it is immutable, we cannot deserialize it anymore?

docs/spring-boot/app-starter.md
symphony-hong: I think it will be useful when user wants to create project not using the generator

docs/spring-boot/app-starter.md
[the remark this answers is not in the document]
symphony-hong: I just realize that the `ExtensionAppAuthenticator` bean (which is in `core-starter`) will be created if the `appId` is defined, so if we want to move `appId` under `bdk-app`, then the `SymphonyBdkAppProperties` must be declared in `core-starter` instead of `app-starter`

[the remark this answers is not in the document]
symphony-thibault: `app-starter` depends on `core-starter`. So `appId` etc. have to be configured at `core-starter` level: 
```yaml
bdk: # configure core
   appId: blabla
bdk-app: # configure circle of trust controller only, among other future things (note that this configuration could be optional)
   auth: 
       enabled: true
```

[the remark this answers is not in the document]
symphony-hong: It has to be:
```
bdk: # configure core
   app:
     appId: blabla
     privateKey: /path/to/privateKey
bdk-app: # configure circle of trust controller only, among other future things (note that this configuration could be optional)
   auth: 
       enabled: true
```
because the coreProperties will be used as the config passed to the [AuthenticatorFactory](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkCoreConfig.java#L63-L67)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/JwtCookieProperties.java
symphony-hong: do you know how can use `@Min` with the `Duration` field?

symphony-thibault: Then maybe a duration cannot be negative, so no need to check `@Min`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/TokenPair.java
symphony-thibault: https://www.baeldung.com/jackson-deserialize-immutable-objects

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/JwtCookieProperties.java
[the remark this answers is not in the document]
symphony-hong: I change the field into `Duration` and I'm not sure how to use `@Min` here

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/TokenPair.java
symphony-thibault: https://www.google.com/search?q=jackson+immutable+objects :)

symphony-hong: OOOOPPPSSS! 😴

docs/spring-boot/app-starter.md
```diff
@@ -0,0 +1,210 @@
...
+    
+    implementation 'com.symphony.platformsolutions:symphony-bdk--spring-boot-starter'
+    implementation 'org.springframework.boot:spring-boot-starter'
+}
+```
+
+## Create a Simple Backend for Extension Application
+As a first step, you have to initialize your environment through the Spring Boot `src/main/resources/application.yaml` file: 
+```yaml
+bdk:
```
symphony-thibault, line 82: We should have: 
```yaml
bdk:
  host: acme.symphony.com
  bot:
    username: mybot
    privateKeyPath: /path/to/rsa/privatekey.pem
  app:
    appId: myapp
    privateKeyPath: /path/to/rsa/privatekey.pem

bdk-app:
  auth:
    enabled: true # enable the CircleOfTrust controller (default is true)
    jwtCookie:
      enabled: true
      maxAge: 1d # see https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration
```
⚠️ You should really consider the documentation as also important as the code you write. If the documentation is wrong, developers won't know how to use what you developed. And they might contact the Symphony support for that, leading to a support request opened on our side and finally time consumed for nothing (for us and support guys).

symphony-thibault: Documentation is not something we write just for fun 

attention

symphony-bdk-examples/bdk-app-spring-boot-example/build.gradle
```diff
@@ -0,0 +1,13 @@
...
+    id 'org.springframework.boot' version "2.3.4.RELEASE"
+}
+
+description = 'Symphony Java BDK Examples for the SpringBoot integration'
+
+dependencies {
+
+    implementation project(':symphony-bdk-spring:symphony-bdk-app-spring-boot-starter')
+
+//    implementation 'org.springframework.boot:spring-boot-starter'
```
symphony-thibault, line 11: Do we need it finally?

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/java/com/symphony/bdk/examples/app/spring/AppSpringApplication.java
symphony-thibault, line 7: `ExtensionAppApplication`

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
```diff
@@ -0,0 +1,20 @@
+
+bdk:
+  host: acme.symphony.com
+  bot:
+    username: fxbot-hong
```
symphony-thibault, line 5: `mybot`

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
```diff
@@ -0,0 +1,20 @@
...
+    privateKeyPath: ${user.home}/.symphony/keys/rsa-private-bot.pem
+  app:
+    appId: testapp-hongle
```
symphony-thibault, line 8: `myapp`

symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
```diff
@@ -0,0 +1,20 @@
...
+    privateKeyPath: ${user.home}/.symphony/keys/app_privatekey.pem
+
+bdk-app:
+  auth:
+    enabled: true
+    jwtCookie:
+      enabled: true
+      expireIn: 43200
```
symphony-thibault, line 16: `expireIn` renamed to `maxAge` right? And using duration, not integer value

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/build.gradle
```diff
@@ -13,12 +13,35 @@ jacocoTestCoverageVerification {
...
+    testCompileOnly 'org.projectlombok:lombok'
+    testAnnotationProcessor 'org.projectlombok:lombok'
+
+    testImplementation 'org.junit.vintage:junit-vintage-engine'
+    testCompile project(':symphony-bdk-core').sourceSets.test.output
+    testImplementation('org.springframework.boot:spring-boot-starter-test') {
+        exclude group: 'org.junit.vintage', module: 'junit-vintage-engine'
+        exclude group: 'junit', module: 'junit'
+    }
+    testImplementation 'org.junit.jupiter:junit-jupiter-api'
```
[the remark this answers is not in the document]
symphony-thibault: @symphony-hong ^^

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/model/BdkAppErrorCode.java
```diff
@@ -0,0 +1,30 @@
+package com.symphony.bdk.app.spring.auth.model;
+
+import org.springframework.http.HttpStatus;
+
+/**
+ * Error Code explains explicitly the error returned by the Extension App Backend.
+ */
+public enum BdkAppErrorCode {
+
+  UNAUTHORIZED(HttpStatus.UNAUTHORIZED, "Failed to authenticate the extension app {appId}."),
```
symphony-thibault, line 10: `AUTH_FAILURE`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/exception/GlobalControllerExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private final SymphonyBdkCoreProperties properties;
+
+  public GlobalControllerExceptionHandler(SymphonyBdkCoreProperties properties) {
+    super();
+    this.properties = properties;
+  }
+
+  @ExceptionHandler(BdkAppException.class)
+  public ResponseEntity<Object> handleUnauthorizedException(Exception e, WebRequest request) {
```
symphony-thibault, line 34: `Exception` could directly be `BdkAppException`

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/AppAuthProperties.java
```diff
@@ -0,0 +1,15 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Configuration Properties for enabling Application Authentication APIs.
+ */
+@Getter
+@Setter
+public class AppAuthProperties {
```
symphony-thibault, line 11: Do we import `spring-boot-configuration-processor` as depdency (cf. https://www.baeldung.com/spring-boot-configuration-metadata#nested-properties)?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/AppAuthProperties.java
```diff
@@ -0,0 +1,15 @@
...
+
+  private Boolean enabled = true;
```
symphony-thibault, line 13: javadoc is also very important on configuration properties fields (cf. https://www.baeldung.com/spring-boot-configuration-metadata#generating-metadata)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/test/java/com/symphony/bdk/app/spring/SymphonyBdkAppAutoConfigurationTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Test
+  void shouldLoadContextWithSuccess() {
+
+    final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+        .withPropertyValues(
+            "bdk.app.appId=test-app",
+            "bdk.app.privateKeyPath=classpath:/privatekey.pem",
+            "bdk-app.auth.enabled=true",
+            "bdk-app.auth.jwtCookies.enabled=true",
+            "bdk-app.auth.jwtCookies.maxAge=86400"
```
symphony-thibault, line 23: Could we use a Duration value here?

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/AppAuthProperties.java
symphony-hong: Yes we have it [here](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/pull/285/files#diff-fe5571c0f96e23877232698ad99f3361df1f1f0d3217b73a87db1b0c57307304R32)

symphony-elias (approved):
LGTM

symphony-thibault (approved):
L G T M 🍾 🍾 🍾 🍾 🍾 🍾 🍾  😄

docs/spring-boot/app-starter.md
```diff
@@ -116,30 +116,95 @@ public class ExtAppSpringApplication {
...
-  
-- Validate jwt: `POST /bdk/v1/app/tokens`
-
-    Parameters:
-    ```json
-    {
-      "jwt": "your-jwt" // Jwt signed by Symphony App Frontend
-    }
-    ``` 
+```yaml
```
symphony-youri, line 116: we could put it in a separate file and load via https://editor.swagger.io/?url=https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore.yaml (but how do we target a specific commit... well i guess it won't change much)

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/auth/CircleOfTrustController.java
```diff
@@ -35,49 +29,35 @@
...
+    circleOfTrustService.validateTokens(tokenPair);
   }
 
   @PostMapping("/jwt")
   public UserId validateJwt(@Valid @RequestBody JwtInfo jwtInfo, HttpServletRequest request,
       HttpServletResponse response) {
     log.debug("Validate the jwt signed by extension app frontend to get the user id");
     String jwt = jwtInfo.getJwt();
-    UserId userId = jwtService.validateJwt(jwt);
+    log.debug("JWT " + jwt);
```
symphony-youri, line 58: isn't a sensitive information to log?

symphony-youri (approved):
Good job, that was a big one!

PR #298, opened 2020-10-22
SDK-161 Java-SDK: Create test coverage for the class clients.symphony.api.PresenceClient

Haiddi-Avencod:
### Ticket
[161](https://perzoinc.atlassian.net/browse/SDK-161?atlOrigin=eyJpIjoiYzA5YWFkMWJkZjc1NGM4ZTg3NTZmNzdlM2Q4ZTliMTYiLCJwIjoiaiJ9)

### Coverage
coverage_161

### Comment
I overloaded the method `stubDelete` by adding `status` as parameter in order to call the stub with the desired status


symphony-bdk-legacy/symphony-api-client-java/src/test/java/it/clients/symphony/api/PresenceClientTest.java
```diff
@@ -2,141 +2,536 @@
...
+
+    try {
+
+      final UserPresence userPresence = presenceClient.getUserPresence(1L, true);
+      assertNotNull(userPresence);
+
+      verifyUserPresence(userPresence, 1L, UserPresenceCategory.AVAILABLE, 1533928483800L);
+
+    } catch (SymClientException e) {
+      fail();
```
symphony-elias, line 52: These catch...fail() statements are not useful. Test will fail if exception is thrown.

Haiddi-Avencod: In the method `getUserPresence` the SymClientException is not catched but thrown, so we have to catch it when we call the method. 
From the start we have been doing it like this. When a method throws an exception, we catch it.

Is it a bad way ?

srodriguesavencod: In my opinion it's good we are testing the success if any excpetion is thrown , the test have to failed because it's not a success

Haiddi-Avencod:
There is a failure after having updated the branch but this failure doesn't appear on my side

symphony-bdk-legacy/symphony-api-client-java/src/test/java/it/clients/symphony/api/PresenceClientTest.java
```diff
@@ -2,141 +2,536 @@
...
+import static org.junit.Assert.*;
 
 public class PresenceClientTest extends BotTest {
 
   private PresenceClient presenceClient;
 
   @Before
   public void initClient() {
-    this.presenceClient = new PresenceClient(this.symBotClient);
+    presenceClient = new PresenceClient(symBotClient);
```
symphony-thibault, line 29: I personally prefer to keep the `this`

symphony-bdk-legacy/symphony-api-client-java/src/test/java/it/clients/symphony/api/PresenceClientTest.java
symphony-thibault: As @symphony-elias said the `try/catch` block is useless. The test method can just `throw SymClientException`. The result will be the same and we will have more details in case of error than just calling `fail()`, that would give no detail regarding what happened.

symphony-thibault: Please apply the same for following test methods, it will also improve readability 👍

symphony-elias (approved):
Looks good

PR #301, opened 2020-10-26
Backporting changes from master/symphony-bdk-legacy to 1.3-rc

symphony-hong:


symphony-thibault (approved):

symphony-bdk-legacy/symphony-bdk-bot-sdk-java/build.gradle
```diff
@@ -1,5 +1,14 @@
+plugins {
```
symphony-youri, line 1: this plugin is already applied at top level, it is weird that we had to apply it here

PR #305, opened 2020-10-27
APP-3142: Adding SignalService feature

symphony-hong:
### Ticket
[APP-3142](https://perzoinc.atlassian.net/browse/APP-3142)

### Description
OboSignalService to enable the Obo service for SignalService.

SignalService exposing all endpoints relate to signals information.

Entry to get SignalService from SymphonyBdk entry point.

PaginatedService for SignalService#listSignals and SignalService#subscribers

### Checklist

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
@@ -1,0 +1,128 @@
+package com.symphony.bdk.core.service.signal;
+
+import com.symphony.bdk.gen.api.model.BaseSignal;
+import com.symphony.bdk.gen.api.model.ChannelSubscriber;
+import com.symphony.bdk.gen.api.model.ChannelSubscriptionResponse;
+import com.symphony.bdk.gen.api.model.Signal;
+
+import org.apiguardian.api.API;
+
+import java.util.List;
+import java.util.stream.Stream;
+
+/**
+ * Service interface exposing OBO-enabled endpoints to manage signals information.
+ */
+@API(status = API.Status.STABLE)
+public interface OboSignalService {
+
+  /**
+   * Lists signals on behalf of the user.
+   * {@link SignalService#listSignals(Integer, Integer)}
+   *
+   * @param skip  The number of signals to skip.
+   * @param limit Maximum number of signals to return. Default is 50, maximum value is 500.
+   * @return List of signals that the user has created and public signals to which they have subscribed.
+   * @see <a href="https://developers.symphony.com/restapi/reference#list-signals">List Signals</a>
+   */
+  List<Signal> listSignals(Integer skip, Integer limit);
+
+  /**
+   * Lists signals on behalf of the user.
+   * {@link SignalService#listSignalsStream(Integer, Integer)}
+   *
+   * @param chunkSize size of elements to retrieve in one call. Optional and defaults to 50.
+   * @param totalSize size of elements to retrieve in one call. Optional and defaults to 50.
+   * @return a {@link Stream} containing the signals.
+   * @see <a href="https://developers.symphony.com/restapi/reference#list-signals">List Signals</a>
+   */
+  Stream<Signal> listSignalsStream(Integer chunkSize, Integer totalSize);
+
+  /**
+   * Gets details about the specified signal.
+   * {@link SignalService#getSignal(String)}
+   *
+   * @param id The id of the signal.
+   * @return Details of the specified signal.
+   * @see <a href="https://developers.symphony.com/restapi/reference#get-signal">Get Signal</a>
+   */
+  Signal getSignal(String id);
+
+  /**
+   * Create a new signal.
+   * {@link SignalService#createSignal(BaseSignal)}
+   *
+   * @param signal  A signal object to be created.
+   * @return A new created signal object.
+   * @see <a href=https://developers.symphony.com/restapi/reference#create-signal">Create Signal</a>
+   */
+  Signal createSignal(BaseSignal signal);
+
+  /**
+   * Update an existing signal.
+   * {@link SignalService#updateSignal(String, BaseSignal)}
+   *
+   * @param id      The id of the signal to be updated.
+   * @param signal  The signal object to be updated.
+   * @return The updated signal.
+   * @see <a href="https://developers.symphony.com/restapi/reference#update-signal">Update Signal</a>
+   */
+  Signal updateSignal(String id, BaseSignal signal);
+
+  /**
+   * Delete an existing signal.
+   * {@link SignalService#deleteSignal(String)}
+   *
+   * @param id  The id of the signal to be deleted.
+   * @see <a href="https://developers.symphony.com/restapi/reference#delete-signal">Delete Signal</a>
+   */
+  void deleteSignal(String id);
+
+  /**
+   * Subscribe a list of users to a signal.
+   * {@link SignalService#subscribeSignal(String, Boolean, List)}
+   *
+   * @param id      The id of the signal to be subscribed.
+   * @param pushed  Prevents the user from unsubscribing from the Signal (only for bulk subscriptions).
+   *                Requires the canManageSignalSubscription entitlement.
+   * @param userIds List of user ids to subscribe to the signal
+   * @return The subscription information.
+   * @see <a href="https://developers.symphony.com/restapi/reference#subscribe-signal">Subscribe Signal</a>
+   */
+  ChannelSubscriptionResponse subscribeSignal(String id, Boolean pushed, List<Long> userIds);
+
+  /**
+   * Unsubscribe a list of users from a signal.
+   * {@link SignalService#unsubscribeSignal(String, List)}
+   *
+   * @param id      The id of the signal to be unsubscribed.
+   * @param userIds The list of user ids to unsubscribe from the signal.
+   * @return The unsubscription information.
+   * @see <a href="https://developers.symphony.com/restapi/reference#unsubscribe-signal">Unsubscribe Signal</a>
+   */
+  ChannelSubscriptionResponse unsubscribeSignal(String id, List<Long> userIds);
+
+  /**
+   * Get the subscribers for a specified signal.
+   * {@link SignalService#subscribers(String, Integer, Integer)}
+   *
+   * @param id      The id of the specified signal.
+   * @param skip    The number of results to skip.
+   * @param limit   The maximum number of subscribers to return. If no value is provided, 100 is the default.
+   * @return List of subscribers of the signal.
+   * @see <a href="https://developers.symphony.com/restapi/reference#subscribers">Subscribers</a>
+   */
+  List<ChannelSubscriber> subscribers(String id, Integer skip, Integer limit);
+
+  /**
+   * Get the subscribers for a specified signal.
+   * {@link SignalService#subscribersStream(String, Integer, Integer)}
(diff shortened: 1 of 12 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
```diff
@@ -0,0 +1,128 @@
...
+   */
+  Signal getSignal(String id);
+
+  /**
+   * Create a new signal.
+   * {@link SignalService#createSignal(BaseSignal)}
+   *
+   * @param signal  A signal object to be created.
+   * @return A new created signal object.
+   * @see <a href=https://developers.symphony.com/restapi/reference#create-signal">Create Signal</a>
```
symphony-youri, line 57: looks like a " is missing here (that seems to break syntax coloring in GH)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Lists signals on behalf of the user.
+   * {@link SignalService#listSignals(Integer, Integer)}
+   *
+   * @param skip  The number of signals to skip.
+   * @param limit Maximum number of signals to return. Default is 50, maximum value is 500.
+   * @return List of signals that the user has created and public signals to which they have subscribed.
+   * @see <a href="https://developers.symphony.com/restapi/reference#list-signals">List Signals</a>
+   */
+  List<Signal> listSignals(Integer skip, Integer limit);
```
symphony-youri, line 28: the method names are not very consistent:
- getSignal
- listSignals
- subscribers

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
```diff
@@ -0,0 +1,163 @@
...
+ * Service class for managing signal information.
+ * <p>
+ * This service is used for listing signals related to the user, get information of a specified signal
+ * or perform some actions related to the signal like:
+ * <p><ul>
+ * <li>Create a signal</li>
+ * <li>Update a signal</li>
+ * <li>Delete a signal</li>
+ * <li>Subscribe or unsubscribe a signal</li>
+ * <li></li>
```
symphony-youri, line 33: empty li?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
```diff
@@ -0,0 +1,163 @@
...
+    return executeAndRetry("subscribers",
+        () -> signalsApi.v1SignalsIdSubscribersGet(authSession.getSessionToken(), id, authSession.getKeyManagerToken(), skip, limit)).getData();
+  }
+
+  /**
+   * {@inheritDoc}
+   * @return
+   */
+  @Override
+  public Stream<ChannelSubscriber> subscribersStream(String id, Integer chunkSize, Integer totalSize) {
```
symphony-youri, line 148: passing nulls when you have no ideas of the chunkSize/totalSize to use is not very pretty in term of API
One solution could to have a Pagination object (that comes with default values)
Or to have different methods with/without pagination

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
```diff
@@ -0,0 +1,163 @@
...
+    return executeAndRetry("unsubscribeSignal",
+        () -> signalsApi.v1SignalsIdUnsubscribePost(authSession.getSessionToken(), id, authSession.getKeyManagerToken(), userIds));
+  }
+
+  /**
+   * {@inheritDoc}
+   * @return
+   */
+  @Override
+  public List<ChannelSubscriber> subscribers(String id, Integer skip, Integer limit) {
```
symphony-youri, line 138: is it normal that we have different style of pagination?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
symphony-hong: I'm following the paginating strategy which was implemented in [MessageService](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java#L369)
But I think having different methods with/without pagination should be consider.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
symphony-hong: I think this one is not really a pagination, just a basic call to the endpoint. In this ticket [APP-3183](https://perzoinc.atlassian.net/browse/APP-3183), we also want to expose the skip and limit param in out apis.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Lists signals on behalf of the user.
+   * {@link SignalService#listSignalsStream(Integer, Integer)}
+   *
+   * @param chunkSize size of elements to retrieve in one call. Optional and defaults to 50.
+   * @param totalSize size of elements to retrieve in one call. Optional and defaults to 50.
```
symphony-thibault, line 35: Same description for those 2 params? Also, why not `skip` and `limit`? Looks confusing

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Stream<Signal> listSignalsStream(Integer chunkSize, Integer totalSize) {
+    PaginatedApi<Signal> api = ((this::listSignals));
+
+    final int actualChunkSize = chunkSize == null ? 50 : chunkSize;
+    final int actualTotalSize = totalSize == null ? 50 : totalSize;
```
symphony-thibault, line 73: Why we don't have the same checks in `listSignals(Integer skip, Integer limit)`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
symphony-hong: I took inspiration from this ticket https://perzoinc.atlassian.net/browse/APP-3183. I don't know why but we want to provide both of default pagination strategy (skip and limit) and our own pagination (chunksize and totalsize).

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
symphony-hong: because we can pass null value to the SignalsApi#v1SignalsListGet to make a call to endpoint without specified skip and limit. The default value will be used in agent side.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
```diff
@@ -0,0 +1,162 @@
...
+  public List<Signal> listSignals(Integer skip, Integer limit) {
+    return executeAndRetry("listSignals",
+        () -> signalsApi.v1SignalsListGet(authSession.getSessionToken(), authSession.getKeyManagerToken(), skip, limit));
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Stream<Signal> listSignalsStream(Integer chunkSize, Integer totalSize) {
```
symphony-thibault, line 69: Annotate params with `@Nullable`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/SignalService.java
```diff
@@ -0,0 +1,162 @@
...
+    final int actualTotalSize = totalSize == null ? 50 : totalSize;
+
+    return new PaginatedService<>(api, actualChunkSize, actualTotalSize).stream();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Signal getSignal(String id) {
```
symphony-thibault, line 82: Annotate `id` param with `@javax.annotation.Nonnull`, same for other methods

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Lists paginated stream of signals on behalf of the user.
+   * {@link SignalService#listSignalsStream(Integer, Integer)}
+   *
+   * @param chunkSize size of elements to retrieve in one call. Optional and defaults to 50.
+   * @param totalSize Total maximum number of signals to return. Optional and defaults to 50.
+   * @return a {@link Stream} containing the signals.
+   * @see <a href="https://developers.symphony.com/restapi/reference#list-signals">List Signals</a>
+   */
+  Stream<Signal> listSignalsStream(@Nullable Integer chunkSize, @Nullable Integer totalSize);
```
symphony-youri, line 51: Adding @nullable is good idea, we could do that for existing methods too

symphony-youri: (like message service)

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/ServiceExampleMain.java
```diff
@@ -37,8 +38,15 @@ public static void main(String[] args) throws Exception {
...
+
     StreamFilter streamFilter = new StreamFilter().addStreamTypesItem(new StreamType().type(StreamType.TypeEnum.IM));
-    List<StreamAttributes> streamsList = bdk.obo(oboSession).streams().listStreams(streamFilter);
+    final List<StreamAttributes> streamsList = bdk.obo(oboSession).streams().listStreams(streamFilter);
     log.info("Streams List: {}", streamsList);
+
+    final List<Signal> signalList = bdk.obo(oboSession).signals().listSignals();
+    log.info("Retrieve {} signals", signalList.size());
+    log.info("First signal:");
+    log.info("Signal name: " + signalList.get(0).getName());
```
symphony-youri, line 49: we are mixing {} and string concat (not really a performance problem here but just to be consistent)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/signal/OboSignalService.java
symphony-hong: I think we can do it in another PR to not make the difference of this PR become huge 😄

symphony-thibault: Could be done as part of https://perzoinc.atlassian.net/browse/APP-3183

symphony-youri (approved):
Once we are feature complete in terms of services we should probably review all the service APIs to make they are consistent and easy to use

PR #310, opened 2020-10-28
Generated POM for legacy bot SDK is missing versions

symphony-youri:
Versions for Spring Boot Actuator and Micrometer were not defined in the
constraints. While it was not breaking the build because transitive
dependencies were used, it was breaking the POM generation, leaving the
versions empty.

A project using the SDK as a dependency would then not get the
transitive dependencies automatically.


symphony-thibault (approved):

symphony-hong (approved):
LGTM

PR #311, opened 2020-10-28
Generated POM for legacy bot SDK is missing versions

symphony-youri:
Versions for Spring Boot Actuator and Micrometer were not defined in the
constraints. While it was not breaking the build because transitive
dependencies were used, it was breaking the POM generation, leaving the
versions empty.

A project using the SDK as a dependency would then not get the
transitive dependencies automatically.

(cherry picked from commit 8e231d4e24b5b6264d8afb15540b3771f6ab082f)


symphony-thibault (dismissed):

symphony-hong (approved):
LGTM

PR #315, opened 2020-10-30
Update UserService according to the changes in swagger spec

symphony-hong:
Update UserService according to this changes in swagger spec.
https://github.com/symphonyoss/symphony-api-spec/commit/df84837f6fa9f31cda9b0a33e3ce86b705c0ee3d#diff-729955798b87f663f4d7efeb58ca0da55bf2bd39f913d5372e0258304e239176L1457

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/OboUserService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/OboUserService.java
@@ -17,16 +17,19 @@ import javax.annotation.Nullable;
 public interface OboUserService {
 
   /**
-   * {@link UserService#searchUserByIds(List, Boolean)}
+   * {@link UserService#searchUserByIds(List, Boolean, Boolean)}
    *
    * @param uidList     List of user ids
    * @param local       If true then a local DB search will be performed and only local pod users will be
    *                    returned. If absent or false then a directory search will be performed and users
    *                    from other pods who are visible to the calling user will also be returned.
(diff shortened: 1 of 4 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/OboUserService.java
```diff
@@ -17,55 +17,70 @@
...
    * @param local       If true then a local DB search will be performed and only local pod users will be
    *                    returned. If absent or false then a directory search will be performed and users
    *                    from other pods who are visible to the calling user will also be returned.
+   * @param active      If not set all user status will be returned,
+   *                    if true all active users will be returned,
+   *                    if false all inactive users will be returned
    * @return Users found by user ids
    * @see <a href="https://developers.symphony.com/restapi/reference#users-lookup-v3">Users Lookup V3</a>
    */
-  List<UserV2> searchUserByIds(@Nonnull List<Long> uidList, @Nonnull Boolean local);
```
symphony-youri, line 29: we lost the @nonull annotation here, is it normal?

symphony-hong: I think @nonnull was added by mistake when the service was developed, that's why I removed it.
https://github.com/symphonyoss/symphony-api-spec/blob/master/pod/pod-api-public.yaml#L1456

symphony-youri (approved):

symphony-youri (approved):

PR #336, opened 2020-11-10
APP-3251: Integrated blast message endpoint

symphony-elias:
### Ticket
[APP-3251](https://perzoinc.atlassian.net/browse/APP-3251)

### Description
Integrated blast message endpoint

### Checklist

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
@@ -29,6 +29,7 @@ import com.symphony.bdk.gen.api.model.StreamAttachmentItem;
 import com.symphony.bdk.gen.api.model.V4ImportResponse;
 import com.symphony.bdk.gen.api.model.V4ImportedMessage;
 import com.symphony.bdk.gen.api.model.V4Message;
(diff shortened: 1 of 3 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -201,29 +203,69 @@ public V4Message send(@Nonnull V4Stream stream, @Nonnull Message message) {
...
 
+  /**
+   * Sends a message to multiple existing streams.
+   *
+   * @param streamIds the list of stream IDs to send the message to
+   * @param message   the message to be sent
+   * @return a {@link V4MessageBlastResponse} object containing the details of the sent messages
+   * @see <a href="https://developers.symphony.com/restapi/reference#create-message-v4">Create Message v4</a>
+   */
+  public V4MessageBlastResponse blastMessage(@Nonnull List<String> streamIds, @Nonnull Message message) {
```
symphony-youri, line 218: I'm wondering about the method name vs the existing methods, we could reuse send() but the returned response is a bit weird and maybe we want to make it very explicit it is a blast message
maybe sendBlast? (just to be more consistent with the others send)

symphony-youri (approved):

PR #340, opened 2020-11-10
APP-3113: Throw exception if multiple implementations found in classpath for http-api and template-api

symphony-elias:
### Ticket
[APP-3113](https://perzoinc.atlassian.net/browse/APP-3113)

### Description
Instead of a warn log, an exception is thrown in `ServiceLookup.singleService(Class<?>)`.

### Checklist

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ServiceLookup.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ServiceLookup.java
@@ -18,19 +18,20 @@ public class ServiceLookup {
    * Load a service implementation class using {@link ServiceLoader}.
    *
    * @return a service implementation class.
+   * @throws IllegalStateException if no implementation or several implementations found in classpath.
    */
   public static <T> T lookupSingleService(Class<T> clz) {
 
     final ServiceLoader<T> classServiceLoader = ServiceLoader.load(clz);
 
     final List<T> services = StreamSupport.stream(classServiceLoader.spliterator(), false)
-            .collect(Collectors.toList());
+        .collect(Collectors.toList());
 
     if (services.isEmpty()) {
       throw new IllegalStateException("No service implementation found in classpath.");
     } else if (services.size() > 1) {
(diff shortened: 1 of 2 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/ServiceLookup.java
```diff
@@ -18,19 +18,20 @@
...
   public static <T> T lookupSingleService(Class<T> clz) {
 
     final ServiceLoader<T> classServiceLoader = ServiceLoader.load(clz);
 
     final List<T> services = StreamSupport.stream(classServiceLoader.spliterator(), false)
-            .collect(Collectors.toList());
+        .collect(Collectors.toList());
 
     if (services.isEmpty()) {
       throw new IllegalStateException("No service implementation found in classpath.");
```
symphony-youri, line 31: one issue I had yesterday when trying the BDK on a new project was that I only added the bdk-core dependencies so I had the _No service implementation found in classpath._ error message 
which was not very helpful

At least it could also tell which implem it was looking for that would have helped me to figure out that an HTTP client impl was missing

And even better the error message could help to understand why the issue happens in the first place:
- suggesting that I add the missing impl dependency
- suggesting that I remove one of them (and listing the ones that are present would help too)

symphony-bdk-examples/bdk-core-examples/build.gradle
```diff
@@ -5,7 +5,9 @@ dependencies {
     implementation project(':symphony-bdk-core')
 
     runtimeOnly project(':symphony-bdk-http:symphony-bdk-http-jersey2')
+    //runtimeOnly project(':symphony-bdk-http:symphony-bdk-http-webclient')
     runtimeOnly project(':symphony-bdk-template:symphony-bdk-template-freemarker')
+    runtimeOnly project(':symphony-bdk-template:symphony-bdk-template-handlebars')
```
symphony-thibault, line 10: To comment back before merging

symphony-youri (approved):
👍 for the error messages

PR #341, opened 2020-11-12
APP-3132: HealtService feature

symphony-hong:
### Ticket
[APP-3132](https://perzoinc.atlassian.net/browse/APP-3132)

### Description
Move SessionService to com.symphony.bdk.core.service.session

Create HealthService under com.symphony.bdk.core.service.health with no retry strategy

HealthService entry from SymphonyBdk entry point

Activate SpringBoot autoconfig for HealthService

### Checklist

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
@@ -1,6 +1,7 @@
 package com.symphony.bdk.core.service.health;
 
 import com.symphony.bdk.core.auth.AuthSession;
+import com.symphony.bdk.core.util.function.SupplierWithApiException;
 import com.symphony.bdk.gen.api.SignalsApi;
 import com.symphony.bdk.gen.api.SystemApi;
 import com.symphony.bdk.gen.api.model.AgentInfo;
@@ -36,11 +37,7 @@ public class HealthService {
    * @see <a href="https://developers.symphony.com/restapi/reference#health-check-v3">Health Check v3</a>
    */
   public V3Health v3HealthCheck() {
-    try {
-      return systemApi.v3Health();
-    } catch (ApiException e) {
-      throw new ApiRuntimeException(e);
-    }
+    return execute(systemApi::v3Health);
   }
 
   /**
@@ -51,11 +48,7 @@ public class HealthService {
    * @see <a href="https://developers.symphony.com/restapi/reference#health-check-extended-v3">Healt Check Extended v3</a>
    */
   public V3Health v3ExtendedHealthCheck() {
-    try {
-      return systemApi.v3ExtendedHealth();
-    } catch (ApiException e) {
-      throw new ApiRuntimeException(e);
-    }
+    return execute(systemApi::v3ExtendedHealth);
   }
 
   /**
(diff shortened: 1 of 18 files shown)


symphony-youri (commented):
Probably out of the scope of this PR but we could also think of exposing a health indicator in the spring boot template? https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#writing-custom-healthindicators
And maybe exposing some of the information as a custom info contributors https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-application-info

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apiguardian.api.API;
+
+/**
+ * Service class for checking health of the Agent server.
+ */
+@API(status = API.Status.STABLE)
+public class HealthService {
+
+  private final SystemApi systemApi;
+  private final SignalsApi signalsApi;
```
symphony-youri, line 21: do we intend to fix the swagger spec to fix that strange thing ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Returns the connectivity status of the Agent services as well as users connectivity.
+   * Available on Agent 2.57.0 and above.
+   *
+   * @return {@link V3Health} the connectivity status of the Agent services as well as users connectivity.
+   * @see <a href="https://developers.symphony.com/restapi/reference#health-check-extended-v3">Healt Check Extended v3</a>
+   */
+  public V3Health v3ExtendedHealthCheck() {
```
symphony-youri, line 53: do we want to expose all of them? like do we need the v2 one for instance?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
symphony-hong: we should open a PLAT ticket to fix this thing. I can open if it's really needed

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
symphony-hong: I think the v2 and the extended v3 is mandatory because the v3 healthcheck is not available on all environment. Not sure if we need the default v3 health check, it only returned the status of the agent server.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/health/HealthService.java
```diff
@@ -0,0 +1,95 @@
...
+   * Available on Agent 2.57.0 and above.
+   *
+   * @return {@link V3Health} the connectivity status of your Agent server.
+   * @see <a href="https://developers.symphony.com/restapi/reference#health-check-v3">Health Check v3</a>
+   */
+  public V3Health v3HealthCheck() {
+    try {
+      return systemApi.v3Health();
+    } catch (ApiException e) {
+      throw new ApiRuntimeException(e);
```
symphony-elias, line 42: Do we want to retry ? I guess so, as we already do that for all other calls. And it would avoid the duplication of `try ... catch(ApiException e) { throw new ApiRuntimeException(e);}`  statements.

symphony-hong: In the description of the ticket, @symphony-thibault noted that we won't implement retry strategy for `HealthService`

symphony-elias: Ok, so at least please factorize the try catch and the conversion into an ApiRuntimeException.

symphony-youri (approved):

PR #348, opened 2020-11-16
APP-3138: Data Limit Exceed validate before sending message

symphony-hong:
### Ticket
[APP-3138](https://perzoinc.atlassian.net/browse/APP-3138)

### Description
Data Limit Exceed validate before sending message.

MessageMLValidator to validate the length of the plain text extracted from the messageML.

If the MessageML cannot be parsed by messageml-util, an MessageValidationException will be thrown.

If the plain text extracted from MessageML has the length greater than 81172, an MessageValidationException will be thrown due to the Data Limit exceeded.

### Dependencies

### Checklist

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
@@ -12,6 +12,7 @@ import com.symphony.bdk.core.service.message.model.Message;
 import com.symphony.bdk.core.service.pagination.PaginatedApi;
 import com.symphony.bdk.core.service.pagination.PaginatedService;
 import com.symphony.bdk.core.service.stream.constant.AttachmentSort;
+import com.symphony.bdk.core.util.MessageMLValidator;
 import com.symphony.bdk.core.util.function.SupplierWithApiException;
 import com.symphony.bdk.gen.api.AttachmentsApi;
 import com.symphony.bdk.gen.api.DefaultApi;
@@ -69,6 +70,7 @@ public class MessageService implements OboMessageService, OboService<OboMessageS
   private final DefaultApi defaultApi;
   private final AuthSession authSession;
   private final TemplateEngine templateEngine;
+  private final MessageMLValidator messageMLValidator;
   private final RetryWithRecoveryBuilder<?> retryBuilder;
 
   public MessageService(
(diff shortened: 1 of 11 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -241,6 +249,11 @@ private V4MessageBlastResponse doSendBlast(List<String> streamIds, Message messa
   }
 
   private Map<String, Object> getForm(Message message) {
+    try {
+      this.messageMLValidator.dataExceededValidate(message);
+    } catch (InvalidInputException | IOException | ProcessingException e) {
```
symphony-elias, line 254: I would prefer to have this try...catch statement directly put in `dataExceededValidate(message)` method to have all the logic contained in one place.

symphony-youri (commented):
Should we implement such feature in the BDK?

One of the concern I have is the dependencies brought by this one (https://search.maven.org/artifact/org.symphonyoss.symphony/messageml/0.9.58/jar)
Guava only is a big jar and we have more (like freemarker...)

While we probably don't want to reimplement parsing in the BDK I also wonder how we will implement such feature in the other BDKs

Is this check already implemented in the agent? Could we just have it there (at the cost of calling the agent with a potential failure)

symphony-hong:
@symphony-youri 

> Should we implement such feature in the BDK?

As it is written [here](https://perzoinc.atlassian.net/wiki/spaces/DevX/pages/1263866448/Investigate+Data+Exceeded+Limit+in+Messages+Plaintext), the bot will receive a DataExceededLimit error if the message limit is exceeded. In the page, it stated that we should do the check on both side (BDK and agent)

> One of the concern I have is the dependencies brought by this one (https://search.maven.org/artifact/org.symphonyoss.symphony/messageml/0.9.58/jar)
Guava only is a big jar and we have more (like freemarker...)

Yep and we also have the snyk issues with this dependency

> While we probably don't want to reimplement parsing in the BDK I also wonder how we will implement such feature in the other BDKs

The MessageParser was implemented in the legacy SDK [here](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/6fd04cd6f09ffa10d8926667cfe350e570bd6bc0/symphony-bdk-legacy/symphony-api-client-java/src/main/java/utils/SymMessageParser.java#L52)

PR #350, opened 2020-11-17
APP-3138: Data Limit Exceed validate before sending message

symphony-hong:
### Ticket
[APP-3138](https://perzoinc.atlassian.net/browse/APP-3138)

### Description
Throw a MessageValidationException if the number of characters of message is exceeded the limitation.

### Dependencies

### Checklist


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedServiceV1.java
```diff
@@ -83,7 +84,7 @@ public void start() throws AuthUnauthorizedException, ApiException {
       do {
         readDatafeed();
       } while (this.started.get());
-    } catch (AuthUnauthorizedException | ApiException | NestedRetryException exception) {
+    } catch (AuthUnauthorizedException | ApiException | NestedRetryException | MessageValidationException exception) {
```
symphony-youri, line 87: should we make it a subclass of ApiException so that we don't to change the code everywhere each time we add a new type?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -264,19 +268,26 @@ private V4MessageBlastResponse doSendBlast(List<String> streamIds, Message messa
...
+          headers,
+          emptyMap(),
+          form,
+          apiClient.selectHeaderAccept("application/json"),
+          apiClient.selectHeaderContentType("multipart/form-data"),
+          new String[0],
+          typeReference
+      ).getData();
+    } catch (ApiException e) {
+      if (e.getMessage().contains(MAXIMUM_CHARACTERS_EXCEEDED_ERROR)) {
```
symphony-youri, line 286: would it be enough to forward the error message and leave it as an API Exception? does the user need to handle it as a different type (i'm not sure you can recover from this)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/MessageServiceTest.java
```diff
@@ -441,6 +444,16 @@ void testDoSend(final BdkMockServer mockServer) throws IOException, ApiException
     assertInvokeApiCalledWithCorrectParams(mockServer, message, Collections.emptyList(), Collections.emptyList());
   }
 
+  @Test
+  void testDoSendLimitExceededMessage() throws ApiException {
+    doThrow(new ApiException(400, "Your message exceeds the maximum number of characters allowed."))
+        .when(agentClient).invokeAPI(any(), any(), any(), any(), any(), any(), any(), any(), any(), any(), any());
+
+    String message = "<messageML>" + StringUtils.repeat('a', 70000) + "</messageML>";
```
symphony-youri, line 452: we don't need such a big message for the test :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-hong: if so, I think we have to change nothing because the API exception will be thrown anyway? maybe we only need to log the error message?

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/MessageServiceTest.java
symphony-hong: I just want to make the ut looks realistic :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
symphony-elias: Another attention point is, we should ensure we do not retry when this error occurs.

symphony-youri: I would remove the log entirely (and the hard coded message error) this way we are not tied to a specific error message (which would be duplicated in every BDK)
It is then up to the user of the BDK to log or not this kind of error

symphony-hong: then we have nothing to do with this ticket :) we can easily close this PR

PR #302, opened 2020-10-27
add option to enable/disable datafeed polling

marcus-secato-symphony:
### Ticket
[APP-3199](https://perzoinc.atlassian.net/browse/APP-3199)

### Description
Add a way to enable/disable datafeed polling in legacy BDK

### Dependencies
N/A

### Checklist
- [ ] Referenced a ticket in the PR title and in the corresponding section
- [ ] Filled properly the description and dependencies, if any
- [ ] Unit tests updated or added
- [ ] Javadoc added or updated
- [ ] Updated the documentation in [docs folder](../docs)

PR #359, opened 2020-11-23
APP-3203: Optional start of DF loop in spring boot starter

symphony-elias:
### Ticket
[APP-3203](https://perzoinc.atlassian.net/browse/APP-3203)

### Description
Made the start of the datafeed loop optional. Enabled by default, can be disabled with property `bdk.datafeed.enabled = false`
FYI In the legacy BDK (see #302) choice was to have property `datafeed.polling.disabled=true` in order to disable the DF loop. I made the choice (can subject to comments and changes) to remove the intermediary `polling` object and put `enabled` instead of `disabled` to avoid using a "negative" property name.

### Checklist

### DIFF
--- a/docs/spring-boot/core-starter.md
+++ b/docs/spring-boot/core-starter.md
@@ -213,6 +213,14 @@ public class CoreServices {
 }
 ```
 
+Unlike subscribing to real time events, using slash commands or activities, solely injecting services does not require
+the datafeed loop to run. If you want to disable the datafeed loop, you can update your `application.yaml` file as follows:
+```yaml
+bdk:
+    datafeed:
+        enabled: false
+```
+
 ## Slash Command
 You can easily register a slash command using the `@Slash` annotation. Note that the `CommandContext` is mandatory to 
 successfully register your command. If not defined, a `warn` message will appear in your application log. Note also that 
--- a/docs/spring-boot/core-starter.md
+++ b/docs/spring-boot/core-starter.md
@@ -221,6 +221,8 @@ bdk:
         enabled: false
 ```
 
+:warning: Disabling the datafeed loop will prevent the use of real time event listeners, of slash commands and activities.
+
 ## Slash Command
 You can easily register a slash command using the `@Slash` annotation. Note that the `CommandContext` is mandatory to 
 successfully register your command. If not defined, a `warn` message will appear in your application log. Note also that 
--- a/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
+++ b/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
@@ -13,6 +13,7 @@ import com.symphony.bdk.spring.events.RealTimeEventsDispatcher;
 import com.symphony.bdk.spring.service.DatafeedAsyncLauncherService;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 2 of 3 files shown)


symphony-thibault (approved):

docs/spring-boot/core-starter.md
```diff
@@ -213,6 +213,14 @@ public class CoreServices {
 }
 ```
 
+Unlike subscribing to real time events, using slash commands or activities, solely injecting services does not require
+the datafeed loop to run. If you want to disable the datafeed loop, you can update your `application.yaml` file as follows:
```
symphony-youri, line 217: so this is maybe how we should call the DataFeedService class @symphony-hong (if it is referred to as the datafeed loop)

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/test/java/com/symphony/bdk/spring/SymphonyBdkAutoConfigurationTest.java
```diff
@@ -143,4 +143,31 @@ void shouldInstantiateLoadBalancedApiClientIfConfigured() {
...
+        .withUserConfiguration(SymphonyBdkMockedConfiguration.class)
+        .withConfiguration(AutoConfigurations.of(SymphonyBdkAutoConfiguration.class));
+
+    contextRunner.run(context -> {
+      assertThat(context).hasSingleBean(SymphonyBdkAutoConfiguration.class);
+
+      final SymphonyBdkCoreProperties config = context.getBean(SymphonyBdkCoreProperties.class);
+      assertThat(config.getAgent().getBasePath()).isEqualTo("https://localhost:443");
+
+      assertThat(context).doesNotHaveBean("datafeedService");
```
symphony-youri, line 167: would it be more robust to test on the class type instead of the name? (in case we rename it/change the bean's name)

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
```diff
@@ -24,6 +25,7 @@
 /**
  * Injection of the {@link DatafeedService} instance into the Spring application context.
  */
+@ConditionalOnProperty(value = "bdk.datafeed.enabled", havingValue = "true", matchIfMissing = true)
 public class BdkDatafeedConfig {
```
symphony-youri, line 29: just found out we have too classes named BdkDatafeedConfig in BDK 2.0 maybe we should rename one?

docs/spring-boot/core-starter.md
symphony-youri: maybe we need to be more explicit about what won't work if you use it? (for instance slash commands?)

symphony-elias: > Unlike subscribing to real time events, using slash commands or activities
This is what won't work if you disable the datafeed loop. Maybe I can come up with a more explicit sentence.

symphony-youri: pwahhhh my bad I commented too quickly

symphony-youri (approved):

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
symphony-thibault: Classes in package `com.symphony.bdk.core.config.model` could be suffixed by `Model`, giving `com.symphony.bdk.core.config.model.BdkDatafeedConfigModel`

PR #370, opened 2020-12-02
APP-3296 Add .devx-cicd/config.yaml file, that builds snapshot artifacts

micaste:
### Ticket
[APP-3296](https://perzoinc.atlassian.net/browse/APP-3296)

### Description
See https://github.com/micaste/symphony-api-client-java/blob/APP-3296/.devx-cicd/config.yaml

Introducing the first steps of the devx-cicd builder

My plan is to keep that work on a separate branch for now, but I will still create PRs for merging on it

### DIFF
--- a/.devx-cicd/config.yaml
+++ b/.devx-cicd/config.yaml
@@ -1,0 +1,12 @@
+build:
+  - image: 'adoptopenjdk/openjdk8:slim'
+    commands:
+      - './gradlew -x test'
+publishSnapshot:
+  - image: 'adoptopenjdk/openjdk8:slim'
+    commands:
+      - './gradlew -PsymphonyRepoUrl=https://repo.symphony.com/artifactory/libs-snapshot-local -PsymphonyRepoUser=$ARTIFACTORY_USERNAME -PsymphonyRepoPassword=$ARTIFACTORY_PASSWORD publish'
+publishStableSnapshot:
+  - image: 'adoptopenjdk/openjdk8:slim'
+    commands:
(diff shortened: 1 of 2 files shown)


.devx-cicd/config.yaml
```diff
@@ -0,0 +1,12 @@
+build:
+  - image: 'adoptopenjdk/openjdk8:slim'
+    commands:
+      - './gradlew -x test'
```
symphony-youri, line 4: not directly related to this change
but using gradlew with docker images (or on demand jenkins slaves) means we always redownload the gradle distribution
we could probably try to cache gradle home  or part of it the same way it is done for .m2 (https://docs.gradle.org/current/userguide/gradle_wrapper.html)

.devx-cicd/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+publishSnapshot:
+  - image: 'adoptopenjdk/openjdk8:slim'
+    commands:
+      - './gradlew -PsymphonyRepoUrl=https://repo.symphony.com/artifactory/libs-snapshot-local -PsymphonyRepoUser=$ARTIFACTORY_USERNAME -PsymphonyRepoPassword=$ARTIFACTORY_PASSWORD publish'
```
symphony-youri, line 8: I think we were trying to avoid leaking the internal artifactory URL in this public Git repo (but this is probably ok)

I wish we could get rid of those property and just `gradle publish` using something similar to the m2/settings.xml file

.devx-cicd/config.yaml
micaste: yes, it's an optimization we should be doing. Would you suggest to pre-build the docker image we use with that cache?

.devx-cicd/config.yaml
micaste: if you're uncomfortable with it, we could add some env variables to devx-cicd project

PR #375, opened 2020-12-07
IDEA warnings

symphony-youri:
Cleaning up warnings and making the tests faster for bdk-core module

### DIFF
--- a/symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/AbstractBotAuthenticatorTest.java
+++ b/symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/AbstractBotAuthenticatorTest.java
@@ -19,6 +19,7 @@ import com.symphony.bdk.http.api.ApiRuntimeException;
 
 import org.junit.jupiter.api.Test;
 
+import javax.annotation.Nonnull;
 import javax.ws.rs.ProcessingException;
 
 public class AbstractBotAuthenticatorTest {
@@ -34,6 +35,7 @@ public class AbstractBotAuthenticatorTest {
     }
 
     @Override
+    @Nonnull
     public AuthSession authenticateBot() throws AuthUnauthorizedException {
       return null;
     }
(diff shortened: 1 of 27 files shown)


symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/AbstractBotAuthenticatorTest.java
```diff
@@ -34,6 +35,7 @@ protected String authenticateAndGetToken(ApiClient client) throws ApiException {
     }
 
     @Override
+    @Nonnull
```
symphony-elias, line 38: Not sure is is relevant here

symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/AbstractExtensionAppAuthenticatorTest.java
```diff
@@ -41,6 +42,7 @@ protected ExtensionAppTokens authenticateAndRetrieveTokens(String appToken) thro
       return null;
     }
 
+    @Nonnull
```
symphony-elias, line 45: This function returns null, is it the annotation relevant?

symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/AbstractOboAuthenticatorTest.java
```diff
@@ -45,6 +46,7 @@ protected String authenticateAndRetrieveAppSessionToken() throws ApiException {
       return null;
     }
 
+    @Nonnull
```
symphony-elias, line 49: Returns null

symphony-thibault (approved):

symphony-bdk-core/src/test/java/com/symphony/bdk/core/auth/impl/AbstractBotAuthenticatorTest.java
symphony-youri: Indeed there is warning because @nonnull is not set for implementations but for tests we have dummy implementations
Not sure if there is way to not return null here but I prefer to avoid warnings in the production code and keep them in the test code

symphony-youri: I used Mockito to prevent that

PR #385, opened 2020-12-14
PLAT-9981: Documented change management policy

symphony-elias:
### Ticket
PLAT-9981

### Description
Added documentation file explaining our change management policy.
Added conditional `WARN` logs in case deprecated configuration fields are used.

### Checklist

### DIFF
--- a/docs/tech/api-change-management.md
+++ b/docs/tech/api-change-management.md
@@ -1,0 +1,15 @@
+# API Change management
+
+Rationale of our change management policy is to be backward compatible on existing features
+(classes, methods or [configuration](../configuration.md) fields) whenever possible.
+
+In case of a non-backward compatible change:
+* we will mark the configuration field,
+class or method as deprecated in a given `x.y.z` version.
+* Version `x.(y + 1).0` will start logging a `WARN` message if the class, method or configuration field is called.
+It will be possible for developers to ignore those warnings through JVM parameter `bdk.warning.mode=none`.
+* Deprecated items will be then removed in `x.(y + 2).0` version *only* if a replacement has been
+documented and is available.
+
+Deprecations and subsequent class, method or field removal will all be documented in
+[release notes](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/releases).
--- a/docs/tech/api-change-management.md
+++ b/docs/tech/api-change-management.md
@@ -9,7 +9,9 @@ In case of a non-backward compatible change:
 * we will mark the configuration field,
 class or method as deprecated in a given `x.y.z` version.
 * Version `x.(y + 1).0` will start logging a `WARN` message if the class, method or configuration field is called.
-It will be possible for developers to ignore those warnings through JVM parameter `bdk.warning.mode=none`.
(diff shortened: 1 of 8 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/DeprecationLogger.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Objects;
+
+@Slf4j
+@API(status = API.Status.INTERNAL)
+/**
+ * Class to log deprecation message except if system property bdk.warning.mode is set to none.
+ */
+public class DeprecationLogger {
+
+  private static final boolean shouldLogDeprecationMessages = !Objects.equals(System.getProperty("bdk.warning.mode"),
```
symphony-youri, line 15: should be named SHOULD_LOG_DEPRECATION_MESSAGES

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/DeprecationLogger.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apiguardian.api.API;
+
+import java.util.Objects;
+
+@Slf4j
+@API(status = API.Status.INTERNAL)
+/**
+ * Class to log deprecation message except if system property bdk.warning.mode is set to none.
+ */
+public class DeprecationLogger {
```
symphony-youri, line 13: can be final and with a private constructor as it is an util class

symphony-youri, line 13: rather than having a specific system property what about relying on the log configuration to enable or disable it?
either we could drop the property and give a meaningful name to the logger
and then you could use 

logging.level.BDK_DEPRECATION=ERROR 

to disable the logs?

(for instance https://www.elastic.co/guide/en/elasticsearch/reference/6.7/logging.html#deprecation-logging)

docs/tech/api-change-management.md
```diff
@@ -0,0 +1,15 @@
+# API Change management
+
+Rationale of our change management policy is to be backward compatible on existing features
+(classes, methods or [configuration](../configuration.md) fields) whenever possible.
```
symphony-youri, line 4: it might be worth linking to https://developers.symphony.com/restapi/docs/api-change-management to explain what we do in terms of REST API that are called by the BDK

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/DeprecationLogger.java
```diff
@@ -0,0 +1,28 @@
...
+
+  private static final boolean shouldLogDeprecationMessages = !Objects.equals(System.getProperty("bdk.warning.mode"),
+      "none");
+
+  /**
+   * Logs a WARN message except if system property bdk.warning.mode=none
+   *
+   * @param message deprecation message to log.
+   */
+  public static void logDeprecation(String message) {
```
symphony-youri, line 23: We should have a unit test for that (ok the code is pretty straightforward but by making it an habit helps for other parts)

symphony-youri (approved):

PR #388, opened 2020-12-16
PLAT-10261: Fixed classpath paths for windows platform

symphony-elias:
### Ticket
PLAT-10261

### Description
Fixed path joining for classpath resources on Windows.

### Checklist

### DIFF
--- a/symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
+++ b/symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
@@ -28,30 +28,33 @@ public class FileHelper {
    */
   @SneakyThrows
   public static byte[] readFile(@Nonnull final String path) throws FileNotFoundException {
-
-    byte[] content;
-
-    if(!isClasspath(path) && new File(path).exists()) {
-      try(FileInputStream fis = new FileInputStream(path)) {
-          content = toByteArray(fis);
-          logger.debug("File loaded from system path : {}", path);
-        }
-        catch (Exception ex) {
-          throw new FileNotFoundException("Unable to load file from path : " + path);
-        }
-    }
-    else if (FileHelper.class.getResource(classpath(path)) != null) {
-      content = toByteArray(FileHelper.class.getResourceAsStream(classpath(path)));
+    if (!isClasspath(path) && new File(path).exists()) {
+      try (FileInputStream fis = new FileInputStream(path)) {
+        logger.debug("File loaded from system path : {}", path);
+        return toByteArray(fis);
+      } catch (Exception ex) {
+        throw new FileNotFoundException("Unable to load file from path : " + path);
+      }
+    } else if (FileHelper.class.getResource(classpath(path)) != null) {
       logger.debug("File loaded from classpath location : {}", path);
+      return toByteArray(FileHelper.class.getResourceAsStream(classpath(path)));
     }
-    else {
-      throw new FileNotFoundException("Unable to load file from path : " + path);
-    }
-    return content;
+    throw new FileNotFoundException("Unable to load file from path : " + path);
   }
 
-  public static String path(String first, String... more) {
-    return Paths.get(first, more).toString();
+  public static String path(String first, String second) {
+    return isClasspath(first) ? classpathPath(first, second) : Paths.get(first, second).toString();
+  }
+
+  private static String classpathPath(String first, String second) {
+    // classpath paths are always with '/'
+    if (first.endsWith("/") && second.startsWith("/")) {
+      return first + second.substring(1);
+    }
+    if (!first.endsWith("/") && !second.startsWith("/")) {
+      return first + "/" + second;
+    }
+    return first + second;
   }
 
   private static String classpath(String path) {
--- a/symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
+++ b/symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
@@ -30,16 +30,14 @@ public class FileHelper {
   public static byte[] readFile(@Nonnull final String path) throws FileNotFoundException {
     if (!isClasspath(path) && new File(path).exists()) {
       try (FileInputStream fis = new FileInputStream(path)) {
-        logger.debug("File loaded from system path : {}", path);
+        logger.debug("File loaded from system path: {}", path);
         return toByteArray(fis);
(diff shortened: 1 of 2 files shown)


symphony-bdk-legacy/symphony-api-client-java/src/test/java/internal/FileHelperTest.java
```diff
@@ -29,4 +29,24 @@ public void should_read_file_from_system() throws IOException {
   public void fail_to_read_file() throws FileNotFoundException {
     FileHelper.readFile(UUID.randomUUID().toString() + ".abc");
   }
-}
\ No newline at end of file
+
```
symphony-youri, line 32: may I suggest a test with just classpath:, /file and classpath:, file
it looks it could be a corner case too

for this kind of test that are very simple and in the form of one input, one output I would sometime have a single test method and all the test cases in the same method
(but I don't mind having separated methods too :))

symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
```diff
@@ -28,30 +28,33 @@
...
-        }
-    }
-    else if (FileHelper.class.getResource(classpath(path)) != null) {
-      content = toByteArray(FileHelper.class.getResourceAsStream(classpath(path)));
+    if (!isClasspath(path) && new File(path).exists()) {
+      try (FileInputStream fis = new FileInputStream(path)) {
+        logger.debug("File loaded from system path : {}", path);
+        return toByteArray(fis);
+      } catch (Exception ex) {
+        throw new FileNotFoundException("Unable to load file from path : " + path);
```
symphony-youri, line 36: just noted that there is an extra space before :

symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
```diff
@@ -28,30 +28,33 @@
...
   }
 
-  public static String path(String first, String... more) {
-    return Paths.get(first, more).toString();
+  public static String path(String first, String second) {
+    return isClasspath(first) ? classpathPath(first, second) : Paths.get(first, second).toString();
+  }
+
+  private static String classpathPath(String first, String second) {
+    // classpath paths are always with '/'
```
symphony-youri, line 50: are you sure? to me classpath:blah.txt is a valid classpath

symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
```diff
@@ -28,30 +28,33 @@
...
-    else if (FileHelper.class.getResource(classpath(path)) != null) {
-      content = toByteArray(FileHelper.class.getResourceAsStream(classpath(path)));
+    if (!isClasspath(path) && new File(path).exists()) {
+      try (FileInputStream fis = new FileInputStream(path)) {
+        logger.debug("File loaded from system path : {}", path);
+        return toByteArray(fis);
+      } catch (Exception ex) {
+        throw new FileNotFoundException("Unable to load file from path : " + path);
+      }
+    } else if (FileHelper.class.getResource(classpath(path)) != null) {
```
symphony-youri, line 38: it is a bit weird that here we use getResource first and then getResourceAsStream (it was like that before)

also to load relative from this class is perhaps a bit weird for the user of the SDKs
perhaps we should just use ClassLoader.getSystemResourceAsStream ? (I started looking into ClassPathResource from Spring and it gets messy quickly, picking the right classloader, forward slash or not...)

symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
symphony-elias: I didn't want to say classpath paths always begin with '/' but rather that it cannot contain '\' like Windows (path separator is always '/' regardless of the OS).

symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
symphony-elias: Yes indeed, in Spring things are messy.
But please check: https://docs.oracle.com/javase/8/docs/technotes/guides/lang/resources.html

> **Using Methods of java.lang.Class**
The getResource and getResourceAsStream methods find a resource with a given name. They return null if they do not find a resource with the specified name. The rules for searching for resources associated with a given class are implemented by the class's ClassLoader. The Class methods delegate to ClassLoader methods, after applying a naming convention: if the resource name starts with "/", it is used as is. Otherwise, the name of the package is prepended, after converting all periods (.) to slashes (/).

So for our usecases, I guess it is enough and will work if paths begin with `classpath:/`. Things get messier if there is no slash after `classpath:`.

symphony-bdk-legacy/symphony-api-client-java/src/test/java/internal/FileHelperTest.java
```diff
@@ -29,4 +30,40 @@ public void should_read_file_from_system() throws IOException {
...
+  }
+
+  @Test
+  public void only_classpath_on_first_slash_on_second() {
+    assertEquals("classpath:/file.pem", FileHelper.path("classpath:", "/file.pem"));
+  }
+
+  @Test
+  public void only_classpath_first_with_no_slash() {
+    assertEquals("classpath:/file.pem", FileHelper.path("classpath:", "file.pem"));
```
symphony-elias, line 66: Please mind that `Paths.get("classpath:", "file.pem")` (=current behavior) will add a path separator in between (`/` on linux/macOS or `\` on Windows)

symphony-youri (approved):

symphony-bdk-legacy/symphony-api-client-java/src/main/java/internal/FileHelper.java
```diff
@@ -28,30 +28,31 @@
...
-      try(FileInputStream fis = new FileInputStream(path)) {
-          content = toByteArray(fis);
-          logger.debug("File loaded from system path : {}", path);
-        }
-        catch (Exception ex) {
-          throw new FileNotFoundException("Unable to load file from path : " + path);
-        }
+  private static String classpathPath(String first, String second) {
+    // Path separator for classpath paths is always '/' regardless of the OS
+    if (first.endsWith("/") && second.startsWith("/")) {
```
symphony-youri, line 49: maybe have a constant for the / ?

PR #401, opened 2021-01-06
PLAT-10144: Adding built-in help command 

symphony-hong:
### Ticket
[PLAT-10144](https://perzoinc.atlassian.net/browse/PLAT-10144)

### Description
Adding built-in help command 

### Dependencies

### Checklist


symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -28,9 +31,37 @@
...
+  public ActivityRegistry(UserV2 botSession, Consumer<RealTimeEventListener> realTimeEventsSource , MessageService messageService) {
+    this.botSession = botSession;
+    this.realTimeEventsSource = realTimeEventsSource;
+    this.messageService = messageService;
+
+    this.registerHelpCommand();
+  }
+
+  private void registerHelpCommand() {
+    this.helpCommand = SlashCommand.slash("/help", commandContext -> {
```
symphony-thibault, line 50: What happens if another `/help` command is registered by bot developer?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -28,9 +31,37 @@
...
+      List<String> commands = new ArrayList<>();
+      for (AbstractActivity<?, ?> activity : activityList) {
+        ActivityInfo info = activity.getInfo();
+        if (info instanceof CommandActivityInfo) {
+          CommandActivityInfo commandInfo = (CommandActivityInfo) info;
+          commands.add("<li>" + commandInfo.commandName() + ": " + commandInfo.summary() + "</li>");
+        }
+      }
+      String message = "<ul>" + String.join("\n", commands) + "</ul>";
```
symphony-thibault, line 59: Use a template if possible

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -28,9 +31,37 @@
...
+
   /** The Datafeed real-time events source, or Datafeed listener */
   private final Consumer<RealTimeEventListener> realTimeEventsSource;
 
+  public ActivityRegistry(UserV2 botSession, Consumer<RealTimeEventListener> realTimeEventsSource , MessageService messageService) {
+    this.botSession = botSession;
+    this.realTimeEventsSource = realTimeEventsSource;
+    this.messageService = messageService;
+
+    this.registerHelpCommand();
```
symphony-thibault, line 46: Pass `MessageService` as `registerHelpCommand` parameter and remove from class properties

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -28,9 +31,37 @@
...
+      for (AbstractActivity<?, ?> activity : activityList) {
+        ActivityInfo info = activity.getInfo();
+        if (info instanceof CommandActivityInfo) {
+          CommandActivityInfo commandInfo = (CommandActivityInfo) info;
+          commands.add("<li>" + commandInfo.commandName() + ": " + commandInfo.summary() + "</li>");
+        }
+      }
+      String message = "<ul>" + String.join("\n", commands) + "</ul>";
+      this.messageService.send(commandContext.getStreamId(), Message.builder().content(message).build());
+    }, "Bdk Help Command");
```
symphony-thibault, line 61: Is the help command listed in... help command result?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/SlashCommand.java
```diff
@@ -36,20 +37,47 @@ public static SlashCommand slash(@Nonnull String slashCommandName, @Nonnull Cons
...
+
+  /**
+   * Returns a new {@link SlashCommand} instance.
+   *
+   * @param slashCommandName   Identifier of the command (ex: '/gif' or 'gif').
+   * @param callback           Callback to be processed when command is detected.
+   * @param summary            The summary of the command.
+   * @return a {@link SlashCommand} instance.
+   */
+  public static SlashCommand slash(@Nonnull String slashCommandName, @Nonnull Consumer<CommandContext> callback, String summary) {
```
symphony-thibault, line 58: To be modified on Spring Boot side as well (`@Slash` annotation)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+@Getter
+@Setter
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class CommandActivityInfo extends ActivityInfo {
+
+  private String commandName;
+
+  private String summary;
```
symphony-thibault, line 21: Difference between summary and description from parent class?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-hong: I didn't think about it. Maybe we should have a Map to store all the commands with the command name?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
symphony-hong: the default slash command has this description: `description("Usage: " + (this.requiresBotMention ? "@BotMention " : "") + this.slashCommandName)`, that's why I thought it has the different use rather than a short description to be shown in the help command result.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-hong: yes it is, do we really need to exclude it?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-thibault: The only issue I see is that we don't know at this level what will be the selected template engine

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
symphony-thibault: No problem to change that 👍

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-thibault: Maybe not, could be done later if necessary

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-thibault: Maybe a more general question: what happens if a developer override an existing slash command already defined in his project? 

I'd say: skip it or crash. But any developer should have the ability to override the `/help` command for sure

symphony-youri: We should let developers override already defined /slash commands IMHO (and even unregister existing ones too in case you don' want the help command)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-youri: I was going to say we should not list it but then I checked what `ls --help` does for instance and it does list the --help arg so I guess it is ok to have it

Bdk-> BDK? (or maybe just `List available commands` ?)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
```diff
@@ -0,0 +1,22 @@
...
+import org.apiguardian.api.API;
+
+/**
+ * {@link com.symphony.bdk.core.activity.command.CommandActivity} information/documentation model.
+ */
+@Getter
+@Setter
+@Accessors(fluent = true)
+@API(status = API.Status.EXPERIMENTAL)
+public class CommandActivityInfo extends ActivityInfo {
```
symphony-youri, line 17: aren't we going into future problems with inheritance (vs interface) for instance being limited by the lack of multiple inheritance?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
```diff
@@ -0,0 +1,22 @@
...
+
+  private String commandName;
```
symphony-youri, line 19: do we use it?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
symphony-youri: I don't think we really use the description so far so we could change its default format (is it what you meant @symphony-thibault ?)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-youri: furthermore to avoid the dependency between ActivityRegistry and MessageService at all I would try to call 
this.activityRegistry.register(new HelpCommand(messageService, activityRegistry)) from the SymphonyBdk constructor

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-youri: maybe just using String.format and building the message in order (that is add the <ul> first then the lines, then </ul>) would make it more template like?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivityInfo.java
symphony-hong: I will reuse the `name` and the `description` of the `ActivityInfo`. So we can completely remove this class

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/HelpCommand.java
```diff
@@ -0,0 +1,70 @@
...
+    return Pattern.compile("^" + botMention + HELP_COMMAND + "$");
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  protected void onActivity(CommandContext context) {
+    List<ActivityInfo> infos = new ArrayList<>();
+    for (AbstractActivity<?, ?> act : this.activityRegistry.getActivityList()) {
```
symphony-elias, line 51: You could simplify that: 
```java
this.activityRegistry.getActivityList()
  .stream()
  .filter(a -> !(act instanceof HelpCommand))
  .map(a -> a.getInfo())
  .filter(info -> info.getType().equals(ActivityType.COMMAND))
  .collect(Collectors.toList());
```

symphony-elias (approved):
Looks good

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -56,6 +57,8 @@
   private final SessionService sessionService;
   private final HealthService healthService;
 
+  private final HelpCommand helpCommand;
```
symphony-youri, line 60: it could be a local variable right?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -96,6 +99,13 @@ protected SymphonyBdk(BdkConfig config, ApiClientFactory apiClientFactory)
     // setup activities
     this.activityRegistry =
```
symphony-youri, line 100: would it not be better to have proper ifs here? to follow the flow
if we have a DF loop then 
     we create a registry
     if we have a message service
        then we add the help command to it

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -42,6 +46,11 @@ public void register(final AbstractActivity<?, ?> activity) {
   }
 
   private void preProcessActivity(AbstractActivity<?, ?> activity) {
+    // check if the activity is a help command or not
+    // If yes, replace the existing one
+    if (activity instanceof HelpCommand && this.activityList.size() > 0) {
+      this.activityList.set(0, activity);
```
symphony-youri, line 52: aren't we going to end up with two help command? (    this.activityList.add(activity); in register ())

this.activityList.size() > 0 is more expressive with !this.activityList.isEmpty() (i.e not empty vs size greater than 0)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/ActivityRegistryTest.java
```diff
@@ -43,9 +46,9 @@ void setUp() {
   @Test
   void shouldRegisterActivity() {
     final CommandActivity<?> act = new TestCommandActivity();
-    assertTrue(this.registry.getActivityList().isEmpty(), "Registry must be empty");
+    assertEquals(0, this.registry.getActivityList().size());
```
symphony-youri, line 49: there is an assertj isEmpty method that is more expressive too

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/ActivityRegistryTest.java
```diff
@@ -43,9 +46,9 @@ void setUp() {
...
     this.registry.register(act);
-    assertEquals(1, this.registry.getActivityList().size(), "Registry must contain only 1 activity");
+    assertEquals(1, this.registry.getActivityList().size(), "Registry must contain 2 activity");
```
symphony-youri, line 51: 2 or 1 ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
```diff
@@ -14,7 +14,7 @@
  */
 @Getter
 @Setter
-@Accessors(fluent = true)
+@AllArgsConstructor
```
symphony-youri, line 17: that's is an API breaking change, the class is marked as experimental but still are we sure about it?

symphony-bdk-core/src/main/resources/template/default_help_command.ftl
```diff
@@ -0,0 +1,7 @@
+<messageML>
```
symphony-youri, line 1: aren't we going to have the issue that @symphony-thibault mentioned? (handlebar being used instead)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-youri: furthermore activityList is exposed (@getter) 
what if I mess up with the list, add my activity in position 0 and then add my custom help command?

symphony-youri: now to solve you might wonder if activity should have a sort of identity to be able to replace activities if we have duplicates
maybe this could be the activity info object? it could work for slash commands but might be limited for other types of commands

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
symphony-hong: because we cannot use the object in freemarker template with the fluent accessors, that's why I removed it. But if we decide not to use template at all, we can revert this change.

symphony-bdk-core/src/main/resources/template/default_help_command.ftl
symphony-hong: I found the `symphony-bdk-template:symphony-bdk-template-freemarker` in the dependencies of the core module, so I guess freemarker used by default in the core

symphony-youri: with a limited scope testRuntimeOnly

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/HelpCommand.java
```diff
@@ -0,0 +1,67 @@
...
+   */
+  @Override
+  protected void onActivity(CommandContext context) {
+    List<ActivityInfo> infos = this.activityRegistry.getActivityList()
+        .stream()
+        .filter(act -> !(act instanceof HelpCommand))
+        .map(AbstractActivity::getInfo)
+        .filter(info -> info.getType().equals(ActivityType.COMMAND))
+        .collect(Collectors.toList());
+    Template template = this.messageService.templates().newTemplateFromClasspath("/template/default_help_command.ftl");
```
symphony-thibault, line 56: What if selected template ending is Handlebars?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
symphony-thibault: I'm ok to modify it as it was marked as experimental

symphony-bdk-core/src/main/resources/template/default_help_command.ftl
symphony-thibault: Developer can exclude this dependency to use `symphony-bdk-template:symphony-bdk-template-handlebars`

symphony-hong: so I guess we should consider not to use template in this case? maybe building a string in order is enough?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
symphony-hong: I just found another issue is that if we mess up with the list, we will also have the conflict between the activity list and the datafeedloop subscribers. So I think we should only return a copy of the activity by the getter, right?

symphony-youri (commented):
Should we document somewhere that this help command exists now? (I'm still wondering if we should enable it by default, it could also be part of the bdk but not added by default, adding it manuallt is quite easy too, as a way to reduce risks)

symphony-youri:
Otherwise it looks ok to me

symphony-thibault (approved):

PR #404, opened 2021-01-06
PLAT-10338 Rename wrong CORS properties names

symphony-thibault:
### Ticket
[PLAT-10338](https://perzoinc.atlassian.net/browse/PLAT-10338)

### Description
Renamed CORS properties: 
- allowed-origin → allowed-origin**s**
- allowed-credentials → **allow**-credentials

A `WARN` message is displayed at application startup if old properties are still configured in `application.yaml`. 

#### Why we did that? 
We initially made 2 small mistakes when naming our CORS properties compared to how they are named in Spring Boot: see [CorsProperties](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/cors/CorsConfiguration.html).

### Checklist

### DIFF
--- a/symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/CorsProperties.java
+++ b/symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/CorsProperties.java
@@ -2,6 +2,7 @@ package com.symphony.bdk.app.spring.properties;
 
 import lombok.Getter;
 import lombok.Setter;
+import org.apiguardian.api.API;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


symphony-elias (approved):
Looks good !

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/CorsProperties.java
```diff
@@ -18,10 +19,17 @@
    */
   private List<String> allowedOrigins = Collections.singletonList("/**");
 
+  /**
+   * [deprecated] use property "allow-credentials" instead
```
symphony-youri, line 23: I think we can also use @deprecated in the Javadoc

symphony-youri (commented):
Should we provide backward compat since it is only deprecated for now?
If I understand correctly here we just warn but won't use the deprecated properties

symphony-bdk-spring/symphony-bdk-app-spring-boot-starter/src/main/java/com/symphony/bdk/app/spring/properties/CorsProperties.java
symphony-thibault: my bad, just updated

symphony-youri (approved):
We should also explain in the commit messages why we made the change (especially here since external people don't have access to Jira)

PR #410, opened 2021-01-12
PLAT-10360: Remove Spring warning during startup

symphony-youri:
### Ticket
PLAT-10360

### Description
When Spring integration is used we could see messages such as:

2021-01-08 11:40:18.767  INFO 86263 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'com.symphony.bdk.spring.config.BdkActivityConfig' of type [com.symphony.bdk.spring.config.BdkActivityConfig] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)

during startup.

This was caused because the BeanPostProcessor for slash commands has a
dependency on ActivityRegistry. This is a bean that has other beans as
dependencies too. So the processor is created after a few beans has
already been created and Spring warns us that those beans might not have
been processed.

We remove this dependency by not injecting the activity registry during
the processor's construction but by retrieving it only when needed (that
is when we start processing slash commands).

I relied on the existing tests and manual testing to check that the
warning is no longer printed.

### Dependencies
N/A

### Checklist
- [X] Filled properly the description and dependencies, if any
- [N/A] Unit tests updated or added
- [N/A] Javadoc added or updated
- [N/A] Updated the documentation in [docs folder](../docs)

### DIFF
--- a/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
+++ b/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
@@ -142,7 +142,7 @@ public class SlashAnnotationProcessor implements BeanPostProcessor, ApplicationC
 
   private void registerSlashMethod(Object bean, Method method, Slash annotation) {
     if (this.registry == null) {
-      registry = applicationContext.getBeanFactory().getBean(ActivityRegistry.class);
+      this.registry = applicationContext.getBeanFactory().getBean(ActivityRegistry.class);
     }
     this.registry.register(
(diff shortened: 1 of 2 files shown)


symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
```diff
@@ -140,6 +141,9 @@ private void processBean(final Object bean, final String beanName, final Class<?
   }
 
   private void registerSlashMethod(Object bean, Method method, Slash annotation) {
+    if (this.registry == null) {
+      registry = applicationContext.getBeanFactory().getBean(ActivityRegistry.class);
```
symphony-elias, line 145: Please be consistent: `this.registry` everywhere or `registry` everywhere.

symphony-elias (approved):
👍

symphony-thibault (approved):

PR #411, opened 2021-01-12
PLAT-10144: Check command list is not empty before sending response

symphony-hong:
### Ticket
[PLAT-10144](https://perzoinc.atlassian.net/browse/PLAT-10144)

### Description
Check command list is not empty before sending response


symphony-elias (approved):

symphony-youri (commented):
should we add a unit test for that?

PR #412, opened 2021-01-12
SDK-188 Java BDK: Create coverage for sse.SsePublisher

Haiddi-Avencod:
### Ticket
[SDK-188](https://perzoinc.atlassian.net/browse/SDK-188?atlOrigin=eyJpIjoiY2JhMTlhZjQ0MTdhNGFhODkyMWZlZjlkMTVjZThmZTAiLCJwIjoiaiJ9)

### Coverage
coverage_188

### Comments
When I tested the method `addSubscriber`, I got a `NullPointerException` because this method used the field private `ConcurrentHashMap<String, List<SseSubscriber>> subscribers` but there was no setter for this field. It was set only in the method `register()` which is private.

So I made a protected setter for this field


symphony-elias (approved):
Looks good !

PR #414, opened 2021-01-14
PLAT-10379: Run Snyk instead of Github integration

symphony-youri:
### Ticket
PLAT-10379

### Description

Two main changes described below.
I have not tested a release (a bit complex) so there a is risk for the next release that we discover issues only when actually releasing

---

CircleCI configuration for Snyk

We want to run Snyk only master branch to avoid exposing the SNYK_TOKEN
on any PR (as the project is OSS).
A separate job is defined for that purpose using the Gradle Snyk plugin.

The SNYK_TOKEN is set in a context that can be reused across DevX jobs.

Also configure an exclusion for log4j that we won't fix.

---

Run Snyk with Gradle + cleaner build …

The Gradle Snyk plugin is used, it adds 2 tasks:
- snyk-test
- snyk-monitor

The task is only configured at the root level and with the option to
look at subproject (it is faster than running it for each project).

Most of the changes for the build are to use convention plugin which is
a better approach to configure builds.

In buildSrc/ we define our own plugins with the common configuration and
we can better compose them by applying then where needed.

For instance we use the publish plugin only on the projects that are
published. The root build is much simpler to read and we avoid the magic
and filtering of subprojects{}

This has not been done for the legacy project as it is ... legacy ...

### Dependencies
N/A

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,22 @@
-# Java Gradle CircleCI 2.0 configuration file
-#
-# Check https://circleci.com/docs/2.0/language-java/ for more details
-#
-version: 2
+version: 2.1
+
+workflows:
+  build-workflow:
+    jobs:
+      - build
+      - snyk:
+          context:
+            - devx
+          filters:
+            branches:
+              only: master
+
 jobs:
   build:
     docker:
       - image: cimg/openjdk:8.0
 
     working_directory: ~/repo
-
-    environment:
 
     steps:
       - checkout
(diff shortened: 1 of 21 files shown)


.circleci/config.yml
```diff
@@ -66,3 +72,29 @@ jobs:
...
+            - gradle-repo-v1-{{ .Branch }}-
+            - gradle-repo-v1-
+
+      - run:
+          name: Snyk 2.0
+          command: ./gradlew snyk-test snyk-monitor
+
+      - run:
+          name: Snyk legacy
+          command: cd symphony-bdk-legacy && ./gradlew snyk-test snyk-monitor
```
symphony-hong, line 100: can we use `./gradlew -p symphony-bdk-legacy snyk-test snyk-monitor` instead of cd?

symphony-youri: I changed it because I noticed the snyk-test plugin was not taking into account the -p option but was running snyk in the current folder

symphony-hong (approved):
LGTM

symphony-thibault (approved):

PR #430, opened 2021-02-09
PLAT-10478  Connection pooling with BDK 2.0

### LINKED ISSUE #429: Connection pooling with BDK 2.0
Hi,
I'm having some issues using the Symphony BDK that I'd like to seek some support with.
In a nutshell, I'm using BDK 2.0 to create a bridge to allow 3rd party tools to send alerts via Symphony to appropriate users/rooms. Its a spring boot app exposing rest end points for services to send alert messages.

All works well for a while, but will fail after a period of time to send messages. Sometimes it can be hours other times weeks.
We do connect through a proxy which may have a part in this, but it seems the issue is around PoolingHttpClientConnectionManager.
I see a debug message stating PoolingHttpClientConnectionManager Connection request [route: {tls}->[proxy server] -> ***.symphony.com:443][Total available: 0; route allocated: 2 of 2; total allocated: 2 of 20]. Seems to suggest that there are no connections available in the pool? Not being released back? Is there perhaps a way I can configure TTL to expire after a few minutes for example to release the connection back to the pool? 

I've seen the issue with SDK versions and now updated to the latest BDK 2.0 in the hope it might be resolved.
AS mentioned, this might be something to do with the proxy server, but thought worth starting a dialogue in case its a known issue?

Is this a common problem and any suggestions to prevent? Any assistance greatly appreciated.

symphony-enrico:
Closes #429 

Cannot reproduce the issue, however the message in the logs:

``
PoolingHttpClientConnectionManager Connection request [route: {tls}->[proxy server] -> ***.symphony.com:443][Total available: 0; route allocated: 2 of 2; total allocated: 2 of 20]
``

means that a connection cannot be allocated because the pool allow only 2 with the same route and they are both used.

The connections are closed when the response is closed so here https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/201859e8b3b8043895f18cbc6aea91c5b5f278e1/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientJersey2.java#L187

As it is inside a finally block, it is unlikely that connection is not released; also, a connection timeout is already set https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/201859e8b3b8043895f18cbc6aea91c5b5f278e1/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java#L51 (with a default value of 15000 ms) and it can be changed.

Most likely the issue (with a random occurrence) is due to more than two threads using the same route for a long time. With this modification, the PoolingHttpClientConnectionManager is created manually, by default with the same number of connection but not limited to two on the same route


symphony-youri:
Does it apply to the other HTTP client that we support (Spring webclient)?

symphony-thibault:
We will also need to make configurable from [`BdkClientConfig.java`](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java) + doc update [here](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/docs/configuration.md#full-configuration-example)

symphony-enrico:
> Does it apply to the other HTTP client that we support (Spring webclient)?

It don't think: Spring Webclient use Netty, that has a default connection pool of 500
https://projectreactor.io/docs/netty/snapshot/reference/index.html#_connection_pool but, most important, the pool should be generic, not with a limited by route, that is the main problem of default PoolingHttpClientConnectionManager (I think that 20 connection in the pool are more than enough...if they are not limited to two for the same route!)

For this reason I don't added the configuration in the generic configuration interface. By the way I set the default max connection value has same value than...default one! So we should not have any difference in memory footprint/performance...the idea is only set max connections per route = max connections

20-goto-10:
Hi Guys,
Thanks for the quick response. 
I did manage to hack around the old SDK version we used before to get the routes allocated to 20, but it quickly gobbled up all 20 connections. As mentioned by @symphony-thibault, having config to control the route allocated and total allocated would be useful. Once these changes are merged, I'll grab a copy and set the allocated routes and total allocated to a higher value than the default of 2/20 respectively and try that.

Incidentally, the response.close(); mentioned is within a try block that does nothing if the close fails.  Not sure I fully follow the comment as to why an exception here doesn't matter; I guess because it goes out of scope and would be gc'd?, but just wonder if a little debug message here might prove useful in the event of an exception?

symphony-enrico:
> Hi Guys,
> Thanks for the quick response.
> I did manage to hack around the old SDK version we used before to get the routes allocated to 20, but it quickly gobbled up all 20 connections. As mentioned by @symphony-thibault, having config to control the route allocated and total allocated would be useful. Once these changes are merged, I'll grab a copy and set the allocated routes and total allocated to a higher value than the default of 2/20 respectively and try that.
> 
> Incidentally, the response.close(); mentioned is within a try block that does nothing if the close fails. Not sure I fully follow the comment as to why an exception here doesn't matter; I guess because it goes out of scope and would be gc'd?, but just wonder if a little debug message here might prove useful in the event of an exception?

The close can fail when response is null (exception during calling) or when it is already closed. I will add a warning log, when the response is not null

20-goto-10:
> > Hi Guys,
> > Thanks for the quick response.
> > I did manage to hack around the old SDK version we used before to get the routes allocated to 20, but it quickly gobbled up all 20 connections. As mentioned by @symphony-thibault, having config to control the route allocated and total allocated would be useful. Once these changes are merged, I'll grab a copy and set the allocated routes and total allocated to a higher value than the default of 2/20 respectively and try that.
> > Incidentally, the response.close(); mentioned is within a try block that does nothing if the close fails. Not sure I fully follow the comment as to why an exception here doesn't matter; I guess because it goes out of scope and would be gc'd?, but just wonder if a little debug message here might prove useful in the event of an exception?
> 
> The close can fail when response is null (exception during calling) or when it is already closed. I will add a warning log, when the response is not null

Just another thought on this, could the finally block be avoided altogether if using try-with-resources where assigning the response which supports AutoCloseable and allow autoclose to look after closing the resource closure. Just means the system looks after tidying up resource closures rather than dealing with in a finally block. This is just pseudo code so may be a reason it wouldn't work or be desirable.

    try (
      Response response = null;
      if ("GET".equals(method)) {
        response = invocationBuilder.get();
      } else if ("POST".equals(method)) {
        response = invocationBuilder.post(entity);
      } else if ("PUT".equals(method)) {
        response = invocationBuilder.put(entity);
      } else if ("DELETE".equals(method)) {
        response = invocationBuilder.method("DELETE", entity);
      } else if ("PATCH".equals(method)) {
        response = invocationBuilder.method("PATCH", entity);
      } else if ("HEAD".equals(method)) {
        response = invocationBuilder.head();
      } else if ("OPTIONS".equals(method)) {
        response = invocationBuilder.options();
      } else if ("TRACE".equals(method)) {
        response = invocationBuilder.trace();
      } else {
        throw new ApiException(500, "unknown method type " + method);
  }
) { 
...
}

symphony-enrico:
> > > Hi Guys,
> > > Thanks for the quick response.
> > > I did manage to hack around the old SDK version we used before to get the routes allocated to 20, but it quickly gobbled up all 20 connections. As mentioned by @symphony-thibault, having config to control the route allocated and total allocated would be useful. Once these changes are merged, I'll grab a copy and set the allocated routes and total allocated to a higher value than the default of 2/20 respectively and try that.
> > > Incidentally, the response.close(); mentioned is within a try block that does nothing if the close fails. Not sure I fully follow the comment as to why an exception here doesn't matter; I guess because it goes out of scope and would be gc'd?, but just wonder if a little debug message here might prove useful in the event of an exception?
> > 
> > 
> > The close can fail when response is null (exception during calling) or when it is already closed. I will add a warning log, when the response is not null
> 
> Just another thought on this, could the finally block be avoided altogether if using try-with-resources where assigning the response which supports AutoCloseable and allow autoclose to look after closing the resource closure. Just means the system looks after tidying up resource closures rather than dealing with in a finally block. This is just pseudo code so may be a reason it wouldn't work or be desirable.
> 
> ```
> try (
>   Response response = null;
>   if ("GET".equals(method)) {
>     response = invocationBuilder.get();
>   } else if ("POST".equals(method)) {
>     response = invocationBuilder.post(entity);
>   } else if ("PUT".equals(method)) {
>     response = invocationBuilder.put(entity);
>   } else if ("DELETE".equals(method)) {
>     response = invocationBuilder.method("DELETE", entity);
>   } else if ("PATCH".equals(method)) {
>     response = invocationBuilder.method("PATCH", entity);
>   } else if ("HEAD".equals(method)) {
>     response = invocationBuilder.head();
>   } else if ("OPTIONS".equals(method)) {
>     response = invocationBuilder.options();
>   } else if ("TRACE".equals(method)) {
>     response = invocationBuilder.trace();
>   } else {
>     throw new ApiException(500, "unknown method type " + method);
> ```
> 
> }
> ) {
> ...
> }

Yes, it is way better and I refactored it

symphony-enrico:
> We will also need to make configurable from [`BdkClientConfig.java`](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java) + doc update [here](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/docs/configuration.md#full-configuration-example)

Ok @symphony-thibault , @20-goto-10  I refactored to make it configurable. BTW they are other configuration parameters already partially implemented but the code was not complete, I implemented them also. Now we have (with their default values):

connectionTimeout: 15000
 readTimeout: 60000
 connectionPoolMax: 20
 connectionPoolPerRoute: 20

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientJersey2.java
```diff
@@ -182,12 +162,29 @@ public ApiClientJersey2(final Client httpClient, String basePath, Map<String, St
             buildResponseHeaders(response),
             respBody);
       }
-    } finally {
-      try {
-        response.close();
```
symphony-elias, line 187: Why should we never close? I understand that in some cases, it does nothing but I'd like to know.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
```diff
@@ -38,26 +35,45 @@ public boolean overridesParentConfig() {
...
   }
 
+  @Override
+  public Integer getConnectionPoolMax() { return thisOrParent(connectionPoolMax, parentConfig::getConnectionPoolMax); }
+
+  @Override
+  public Integer getConnectionPoolPerRoute() { return thisOrParent(connectionPoolPerRoute, parentConfig::getConnectionPoolPerRoute); }
+
+  @Override
+  public BdkProxyConfig getProxy() { return thisOrParent(proxy, parentConfig::getProxy); }
```
symphony-thibault, line 75: Our [`.editorconfig`](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/blob/master/.editorconfig#L133) should prevent for simple methods in 1 line

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientJersey2.java
symphony-enrico: @symphony-elias I used try -with-resources statement, so it is closed automatically (and btw a null check is performed before...so no risk of null pointer as it was before, even if silently intercepted)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/client/ApiClientFactory.java
```diff
@@ -49,16 +58,16 @@ public ApiClientFactory(@Nonnull BdkConfig config, @Nonnull ApiClientBuilderProv
    * @return a new {@link ApiClient} instance.
    */
   public ApiClient getLoginClient() {
-    return buildClient(this.config.getPod(), "/login");
+    return buildClient(LOGIN_CONTEXT_PATH, this.config.getPod());
   }
 
   /**
-   * Returns a fully initialized {@link ApiClient} for Pod API.
+   * Returns a fully initialized {@lite: 500nk ApiClient} for Pod API.
```
symphony-thibault, line 65: `lite: 500nk`? :)

symphony-enrico: 😄 really strange...fixed

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkClientConfig.java
symphony-enrico: changed, but they was a copy-paste of methods already formatted in this way. I will push it later

symphony-youri (commented):
We probably have existing tests to complete for these new properties too

symphony-youri (approved):

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientJersey2.java
```diff
@@ -187,18 +167,37 @@ public ApiClientJersey2(final Client httpClient, String basePath, Map<String, St
...
+        case HttpMethod.HEAD:
+          return invocationBuilder.head();
+        case HttpMethod.OPTIONS:
+          return invocationBuilder.options();
+        case "TRACE":
+          return invocationBuilder.trace();
+        default:
+          throw new ApiException(500, "unknown method type " + method);
+      }
     } catch (ProcessingException e) {
```
symphony-enrico, line 195: @symphony-mariacristina can you check that it is was expected? Because to solve the conflict with your last commit I modified a lot...

symphony-thibault (approved):
Good job @symphony-enrico 👍

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientJersey2.java
symphony-mariacristina: @symphony-enrico I made some tests and the ConnectTimeoutException is thrown by the invocationBuilder so we should be good with this refactoring. 👍

20-goto-10:
Hi guys,
Sorry to keep on, but was just reading a little on PoolingHttpClientConnectionManager and it seems that in order to ensure resources are released FULLY so that the connection can be released back to the pool, its a good idea to call EntityUtils.consume(response.getEntity()); Would it be an idea to make this call at the end of the try block for the response? Just belts and braces? It seems to touch on the problem that I am seeing where connections in the pool drop to 0.
References:
http://hc.apache.org/httpcomponents-client-ga/quickstart.html
https://stackoverflow.com/questions/15969037/why-did-the-author-use-entityutils-consumehttpentity

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -176,6 +200,12 @@ private ClientConfig createClientConfig() {
     clientConfig.property(HttpUrlConnectorProvider.SET_METHOD_WORKAROUND, true);
     // turn off compliance validation to be able to send payloads with DELETE calls
     clientConfig.property(ClientProperties.SUPPRESS_HTTP_COMPLIANCE_VALIDATION, true);
+    // By default PoolingHttpClientConnectionManager, if not configured, has 20 connection in the
+    // pool BUT only 2 max connection per route.
+    PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
```
20-goto-10, line 205: Any chance we could use the PoolingHttpClientConnectionManager with the TTL params and a config to allow setting? This may also help by allowing how long a connection can live.
https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#PoolingHttpClientConnectionManager(long,%20java.util.concurrent.TimeUnit)

symphony-enrico:
> orry to keep on, but was just reading a little on PoolingHttpClientConnectionManager and it seems that in order to ensure resources are released FULLY so that the connection can be released back to the pool, its a good idea to call EntityUtils.consume(response.getEntity()); Would it be an idea to make this call at the end of the try block for the response? Just belts and braces? It seems to touch on the problem that I am seeing where connections in the pool drop to 0.
> References:

EntityUtils.consume() is for closing HttpEntity. However, when using Jersey Invocation builder, normally you need just close Response as far I understand

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-enrico: The PR is already merged and we are releasing a new version. Maybe we can discuss later if you have still problems...

yinan-symphony: keep-alive configuration is also a feature we EIS teams needs. looking forward to having such feature soon

symphony-thibault: Could you please create a new [Github issue](https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/issues/new/choose)?

yinan-symphony: @symphony-thibault ,  thx,  and here it is https://perzoinc.atlassian.net/browse/PLAT-10523

PR #431, opened 2021-02-09
PLAT-10110 Catch exceptions on retry based on ApiClient

symphony-mariacristina:
!!! **Update on last commit** !!!
When a proxy is configured the jersey client is throwing ProcessingException enclosing a ConnectTimeoutException so also in this case I transformed that into a SocketTimeoutException so that we trigger the retry.

### Ticket
https://perzoinc.atlassian.net/browse/PLAT-10110

### Description
Depending on the http client used, network issues can throw different kind of exceptions in case of "connection timeout" and "connection refused" :

**Jersey2 client**
- ProcessingException enclosing a ConnectException
- ProcessingException enclosing a SocketTimeoutException

**Webclient**
- WebClientRequestException enclosing a ConnectException
- WebClientRequestException enclosing a ConnectTimeoutException

This is a draft PR to get some feedbacks on my idea of instead of basing the timeout on the external exception (ProcessingException), to check directly the cause of it. In this case we can limit the retry to only these two conditions (timeout and connection refused).
Only "big change" is that in order to avoid catching two types of exceptions for the timeout (one of it being specific to netty package), inside the webclient I changed the type of the exception from ConnectTimeoutException to SocketTimeoutException and forward the latter. In this way we only deal with exceptions types: ConnectException and SocketTimeoutException.

### Dependencies
No dependency on other PRs

### Checklist

### DIFF
--- a/symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
+++ b/symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
@@ -22,6 +22,7 @@ import org.springframework.util.MultiValueMap;
 import org.springframework.web.reactive.function.BodyInserters;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.WebClient;
+import org.springframework.web.reactive.function.client.WebClientRequestException;
 import reactor.core.Exceptions;
 import reactor.core.publisher.Mono;
 
@@ -146,7 +147,9 @@ public class ApiClientWebClient implements ApiClient {
       if (unwrap instanceof ApiException) {
         throw (ApiException) unwrap;
       } if(e.getCause() instanceof ConnectTimeoutException){
-        throw new RuntimeException(new SocketTimeoutException(e.getMessage()));
+        WebClientRequestException exception = (WebClientRequestException) e;
+        throw new WebClientRequestException(new SocketTimeoutException(e.getMessage()), exception.getMethod(),
+            exception.getUri(), exception.getHeaders());
       } else {
         throw e;
       }
--- a/symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
+++ b/symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
@@ -146,7 +146,8 @@ public class ApiClientWebClient implements ApiClient {
       Throwable unwrap = Exceptions.unwrap(e);
       if (unwrap instanceof ApiException) {
         throw (ApiException) unwrap;
-      } if(e.getCause() instanceof ConnectTimeoutException){
+      }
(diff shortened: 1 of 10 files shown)


symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
```diff
@@ -146,7 +147,9 @@ public ApiClientWebClient(final WebClient webClient, String basePath, Map<String
       if (unwrap instanceof ApiException) {
         throw (ApiException) unwrap;
       } if(e.getCause() instanceof ConnectTimeoutException){
```
symphony-youri, line 149: formatting is off (missing a new line and spaces)

symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
```diff
@@ -146,7 +147,9 @@ public ApiClientWebClient(final WebClient webClient, String basePath, Map<String
...
-        throw new RuntimeException(new SocketTimeoutException(e.getMessage()));
+        WebClientRequestException exception = (WebClientRequestException) e;
+        throw new WebClientRequestException(new SocketTimeoutException(e.getMessage()), exception.getMethod(),
```
symphony-youri, line 151: should it end up in a ConnectException?

symphony-mariacristina: No actually the ConnectTimeoutException corresponds to a "timeout" therefore SocketTimeoutException. If "connection refused" it will throw WebClientRequestException enclosing a ConnectException which is already good.

symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
symphony-mariacristina: fixing it

symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
```diff
@@ -146,7 +147,9 @@ public ApiClientWebClient(final WebClient webClient, String basePath, Map<String
       if (unwrap instanceof ApiException) {
         throw (ApiException) unwrap;
       } if(e.getCause() instanceof ConnectTimeoutException){
-        throw new RuntimeException(new SocketTimeoutException(e.getMessage()));
+        WebClientRequestException exception = (WebClientRequestException) e;
```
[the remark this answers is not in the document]
symphony-mariacristina: ConnectTimeoutException should be always encapsulated by a WebClientRequestException in the webclient but just to avoid issues I will add a check in the if statement.

symphony-youri (approved):

symphony-thibault (approved):

symphony-elias (approved):

PR #446, opened 2021-02-16
PLAT-10446 Adding Convenience/Utils classes

symphony-mariacristina:
### Ticket
[PLAT-10446](https://perzoinc.atlassian.net/browse/PLAT-10446)

### Description
Adding some util classes on message processing to:
- Extract entities from a given incoming message (mentions, hashtags, cashtags, emojis)
- Pre-process an outgoing message by "cleaning" the text message, escaping all special characters that will violate messageML format
- Convert a stream id into a url safe id and viceversa

### Dependencies
No dependency

### Checklist

### DIFF
--- a/symphony-bdk-core/build.gradle
+++ b/symphony-bdk-core/build.gradle
@@ -41,7 +41,6 @@ dependencies {
 
     api 'org.apiguardian:apiguardian-api'
     implementation 'org.slf4j:slf4j-api'
-    implementation 'commons-codec:commons-codec'
     implementation 'commons-io:commons-io'
     implementation 'org.apache.commons:commons-lang3'
     implementation 'org.apache.commons:commons-text'
--- a/symphony-bdk-core/build.gradle
+++ b/symphony-bdk-core/build.gradle
@@ -41,6 +41,7 @@ dependencies {
 
     api 'org.apiguardian:apiguardian-api'
     implementation 'org.slf4j:slf4j-api'
+    implementation 'commons-codec:commons-codec'
     implementation 'commons-io:commons-io'
     implementation 'org.apache.commons:commons-lang3'
     implementation 'org.apache.commons:commons-text'
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
@@ -35,7 +35,6 @@ public final class MessageParser {
    */
   public static List<Long> getMentions(V4Message message) throws MessageParserException {
     List<String> mentionsList = getTags(message, EntityTypeEnum.MENTION);
-    if (mentionsList.isEmpty()) { return Collections.emptyList(); }
     return mentionsList.stream().map(Long::parseLong).collect(Collectors.toList());
   }
 
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
@@ -1,0 +1,138 @@
+package com.symphony.bdk.core.service.message.util;
+
+import com.symphony.bdk.core.service.message.exception.MessageParserException;
+import com.symphony.bdk.gen.api.model.V4Message;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.extern.slf4j.Slf4j;
+import org.apiguardian.api.API;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Helper class for extracting entities inside an incoming {@link com.symphony.bdk.gen.api.model.V4Message} such as:
+ * Mentions, Hashtags, Cashtags, Emojis.
+ */
+@Slf4j
+@API(status = API.Status.EXPERIMENTAL)
+public class MessageParser {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private MessageParser() {
+  }
+
+  /**
+   * Parse data inside the message and returns a list containing the user ids corresponding to the users mentioned
+   *
+   * @param message incoming V4 message to be parsed
+   * @return list of users ids that has been mentioned inside the message
+   */
+  public static List<Long> getMentions(V4Message message) throws MessageParserException {
+    List<String> mentionsList = getTags(message, EntityTypeEnum.MENTION);
+    return mentionsList.stream().map(Long::parseLong).collect(Collectors.toList());
+  }
+
+  /**
+   * Parse data inside the message and returns a list containing the text of the hashtags found
+   *
+   * @param message incoming V4 message to be parsed
+   * @return list of hashtags contained in the message
+   */
+  public static List<String> getHashtags(V4Message message) throws MessageParserException {
+    return getTags(message, EntityTypeEnum.HASHTAG);
+  }
+
+  /**
+   * Parse data inside the message and returns a list containing the text of the cashtags found
+   *
+   * @param message incoming V4 message to be parsed
+   * @return list of cashtags contained in the message
+   */
+  public static List<String> getCashtags(V4Message message) throws MessageParserException {
+    return getTags(message, EntityTypeEnum.CASHTAG);
+  }
+
+  /**
+   * Parse data inside the message and returns a map containing the list of emojis found.
+   * Key of the map are the annotation used to identify the emoji and the values are the their unicode.
+   *
+   * @param message incoming V4 message to be parsed
+   * @return map of emojis contained in the message
+   */
+  public static Map<String, String> getEmojis(V4Message message) throws MessageParserException {
+    return getEmojisList(message);
+  }
+
+  private static List<String> getTags(V4Message message, EntityTypeEnum type) throws MessageParserException {
+    String data = message.getData();
+    if (data == null || data.isEmpty()) {
+      log.info("Data payload of message with id {} is empty, no entities found.", message.getMessageId());
+      return Collections.emptyList();
+    }
+    JsonNode dataJson = getJsonNode(data);
+    List<String> tags = new ArrayList<>();
+    for (JsonNode node : dataJson) {
+      if (isType(type, node)) {
+        tags.add(node.get("id").get(0).get("value").asText());
+      }
+    }
+    return tags;
+  }
+
+  private static Map<String, String> getEmojisList(V4Message message) throws MessageParserException {
+    String data = message.getData();
+    if (data == null || data.isEmpty()) {
+      log.info("Data payload of message with id {} is empty, no entities found.", message.getMessageId());
+      return Collections.emptyMap();
+    }
+    JsonNode dataJson = getJsonNode(data);
+    Map<String, String> emojisMap = new HashMap<>();
+    for (JsonNode node : dataJson) {
+      if (isType(EntityTypeEnum.EMOJI, node)) {
+        emojisMap.put(node.get("data").get("annotation").asText(), node.get("data").get("unicode").asText());
+      }
+    }
+    return emojisMap;
+  }
+
+  private static JsonNode getJsonNode(String data) throws MessageParserException {
+    JsonNode dataJson;
+    try {
+      dataJson = MAPPER.readTree(data);
+    } catch (JsonProcessingException e) {
+      throw new MessageParserException("Failed to extract payload from message data", e);
+    }
+    return dataJson;
+  }
+
+  private static boolean isType(EntityTypeEnum type, JsonNode node) {
+    return type.getValue().equals(node.get("type").asText());
+  }
+
+  @API(status = API.Status.EXPERIMENTAL)
+  public enum EntityTypeEnum {
+    HASHTAG("org.symphonyoss.taxonomy"),
+    CASHTAG("org.symphonyoss.fin.security"),
+    MENTION("com.symphony.user.mention"),
+    EMOJI("com.symphony.emoji");
+
+    private final String value;
+
+    EntityTypeEnum(String value) {
+      this.value = value;
+    }
+
+    public String getValue() {
+      return value;
+    }
+
+  }
+}
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
@@ -6,7 +6,8 @@ import com.symphony.bdk.gen.api.model.V4Message;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import lombok.extern.slf4j.Slf4j;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
 import org.apiguardian.api.API;
 
 import java.util.ArrayList;
@@ -20,14 +21,11 @@ import java.util.stream.Collectors;
  * Helper class for extracting entities inside an incoming {@link com.symphony.bdk.gen.api.model.V4Message} such as:
  * Mentions, Hashtags, Cashtags, Emojis.
  */
-@Slf4j
 @API(status = API.Status.EXPERIMENTAL)
-public class MessageParser {
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class MessageParser {
 
   private static final ObjectMapper MAPPER = new ObjectMapper();
-
-  private MessageParser() {
-  }
 
   /**
    * Parse data inside the message and returns a list containing the user ids corresponding to the users mentioned
@@ -37,6 +35,7 @@ public class MessageParser {
    */
   public static List<Long> getMentions(V4Message message) throws MessageParserException {
     List<String> mentionsList = getTags(message, EntityTypeEnum.MENTION);
+    if (mentionsList.isEmpty()) { return Collections.emptyList(); }
     return mentionsList.stream().map(Long::parseLong).collect(Collectors.toList());
   }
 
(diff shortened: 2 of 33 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Parse data inside the message and returns a map containing the list of emojis found.
+   * Key of the map are the annotation used to identify the emoji and the values are the their unicode.
+   *
+   * @param message incoming V4 message to be parsed
+   * @return map of emojis contained in the message
+   */
+  public static Map<String, String> getEmojis(V4Message message) throws MessageParserException {
+    return getEmojisList(message);
```
symphony-elias, line 71: Could `getEmojisList` be inlined and removed as called only once here?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private static String getReplacementOf(String group) {
+    String replacement = tokens.get(group);
+    if (replacement == null) {
+      replacement = tokens.get("\\" + group);
+    }
+    return replacement;
+  }
+
+  private static Map<String, String> initializeMap() {
```
symphony-elias, line 47: - Is there any third party library that could do that for us ?
- Why do we have double slashes ? I don't really understand. Do we expect our users to add double slashes before the special characters below ? As a developer, I would expect being able to put my raw string without "escaping" slashes.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+    return dataJson;
+  }
+
+  private static boolean isType(EntityTypeEnum type, JsonNode node) {
+    return type.getValue().equals(node.get("type").asText());
+  }
+
+  @API(status = API.Status.EXPERIMENTAL)
+  public enum EntityTypeEnum {
```
symphony-thibault, line 121: Is it something we really want to expose outside of the `MessageParser`?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
```diff
@@ -0,0 +1,138 @@
...
+ * Mentions, Hashtags, Cashtags, Emojis.
+ */
+@Slf4j
+@API(status = API.Status.EXPERIMENTAL)
+public class MessageParser {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  private MessageParser() {
+  }
```
symphony-thibault, line 30: Lombok style `@NoArgsConstructor(access = AccessLevel.PRIVATE)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
```diff
@@ -0,0 +1,138 @@
...
+   * @return map of emojis contained in the message
+   */
+  public static Map<String, String> getEmojis(V4Message message) throws MessageParserException {
+    return getEmojisList(message);
+  }
+
+  private static List<String> getTags(V4Message message, EntityTypeEnum type) throws MessageParserException {
+    String data = message.getData();
+    if (data == null || data.isEmpty()) {
+      log.info("Data payload of message with id {} is empty, no entities found.", message.getMessageId());
```
symphony-thibault, line 77: I'd maybe decrease the logging level to `debug` as it is something correct to not have any tags

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
```diff
@@ -0,0 +1,45 @@
...
+ * Helper class providing stream id conversion
+ */
+@API(status = API.Status.EXPERIMENTAL)
+public class StreamUtil {
+
+  private StreamUtil() {
+  }
+
+  /**
+   * Convert the stream id to the corresponding URLSafe encoded stream id
```
symphony-thibault, line 17: Please give a conversion example in Javadoc (using `<pre>...</pre>`)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/util/MessageParserTest.java
```diff
@@ -0,0 +1,134 @@
...
+        + "  \"4\": {\n"
+        + "    \"id\": [\n"
+        + "      {\n"
+        + "        \"type\": \"com.symphony.user.userId\",\n"
+        + "        \"value\": \"1305690252351\"\n"
+        + "      }\n"
+        + "    ],\n"
+        + "    \"type\": \"com.symphony.user.mention\"\n"
+        + "  }\n"
+        + "}";
```
symphony-thibault, line 132: Would be better to have a JSON file in https://github.com/SymphonyPlatformSolutions/symphony-api-client-java/tree/master/symphony-bdk-core/src/test/resources/message

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
symphony-youri: if it is the expected behavior I would not even log (and often we don't want to log from utilities, it is up to the caller to decide to log or not)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
```diff
@@ -0,0 +1,138 @@
...
+      return Collections.emptyList();
+    }
+    JsonNode dataJson = getJsonNode(data);
+    List<String> tags = new ArrayList<>();
+    for (JsonNode node : dataJson) {
+      if (isType(type, node)) {
+        tags.add(node.get("id").get(0).get("value").asText());
```
symphony-youri, line 84: should we be extra defensive here and check if "id" exists and it it has one element? and if it has a value

there is also .path("") that is a safe navigation method (compared to get)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
symphony-youri: yep, let's make it private maybe

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Helper class for pre-processing an outgoing message
+ */
+@API(status = API.Status.EXPERIMENTAL)
+public class MessageUtil {
+  private static Map<String, String> tokens = initializeMap();
+  private static final Pattern pattern = initializePattern(tokens);
```
symphony-youri, line 17: if it is a constant it should be named PATTERN
it tokens a constant too?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
symphony-youri: maybe StringEscapeUtils ? from apache commons text

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
```diff
@@ -0,0 +1,45 @@
...
+   *
+   * @param streamId of the stream to be parsed
+   * @return stream id after conversion
+   */
+  public static String toUrlSafeStreamId(String streamId) {
+    if (streamId == null || streamId.isEmpty()) {
+      return "";
+    }
+    return streamId.trim()
```
symphony-youri, line 26: don't we want to use the same logic as https://github.com/SymphonyOSF/SymphonyLibrary/blob/master/foundation/src/main/java/com/symphony/streams/StreamId.java#L58
(to avoid regex or at least rely on some existing library)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Convert the stream id to the corresponding URLSafe encoded stream id
+   *
+   * @param streamId of the stream to be parsed
+   * @return stream id after conversion
+   */
+  public static String toUrlSafeStreamId(String streamId) {
+    if (streamId == null || streamId.isEmpty()) {
```
symphony-youri, line 23: i would be ok to throw an NPE is null is passed, then we can remove the isEmpty check too

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Convert the URLSafe encoded stream id to the corresponding original stream id
+   *
+   * @param streamId of the stream to be parsed
+   * @return stream id after conversion
+   */
+  public static String fromUrlSafeStreamId(String streamId) {
+    if (streamId == null || streamId.isEmpty()) {
```
symphony-youri, line 39: same here for null handling

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/util/MessageParserTest.java
```diff
@@ -0,0 +1,134 @@
...
+  @Test
+  public void testNullData() throws MessageParserException {
+    when(message.getData()).thenReturn(null);
+    assertTrue(MessageParser.getMentions(message).isEmpty());
+    assertTrue(MessageParser.getEmojis(message).isEmpty());
+  }
+
+  @Test
+  public void testJsonProcessingException() {
+    when(message.getData()).thenReturn("Unparsable json");
```
symphony-youri, line 73: with junit 5 you can use Assertions.assertThrows

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/util/MessageUtilTest.java
```diff
@@ -0,0 +1,35 @@
+package com.symphony.bdk.core.service.message.util;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class MessageUtilTest {
```
symphony-youri, line 7: parameterized tests could be an option here too
(but I'm ok with the ones below, it is easy to understand)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
symphony-mariacristina: making it private and with  @API(status = API.Status.INTERNAL)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
symphony-mariacristina: added in each util class

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/util/MessageParserTest.java
symphony-mariacristina: sure I will add it there!

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
symphony-mariacristina: you are right! Changed to use the same logic in there.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
symphony-mariacristina: Adding some checks on this, and also unit tests for this scenario.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
symphony-mariacristina: We have all those double slashes to build the regex for the pattern otherwise some of them will be considered as part of the regex and not matched (like: []{}.*). But at the end the developer doesn't need to add the slashes in the raw text.

I tried to use the StringEscapeUtils to escape characters but is not working for all of them. Also I was not able to find a full list of which chars we should really escape that's why I basically kept same behaviour has it was done for the legacy BDK. Only information I found is here: https://developers.symphony.com/symphony-developer/docs/messagemlv2#root-tag, but this is not the full list.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
symphony-mariacristina: done in the method definition for both conversions functions

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
```diff
@@ -0,0 +1,42 @@
+package com.symphony.bdk.core.service.stream.util;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.apache.commons.codec.binary.Base64;
```
symphony-youri, line 5: Would it be possible to use Base64 from the JDK? (just wondering as a way to remove the commons code dependency)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Parse data inside the message and returns a list containing the user ids corresponding to the users mentioned
+   *
+   * @param message incoming V4 message to be parsed
+   * @return list of users ids that has been mentioned inside the message
+   */
+  public static List<Long> getMentions(V4Message message) throws MessageParserException {
+    List<String> mentionsList = getTags(message, EntityTypeEnum.MENTION);
+    if (mentionsList.isEmpty()) { return Collections.emptyList(); }
```
symphony-youri, line 38: we can remove the isempty check, the stream will work on an empty list, returning an empty list

symphony-mariacristina: uh right, I forgot to remove it.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * This method takes care of all special characters placed within the messageML that must be HTML-escaped
+   * to have a valid MessageML format.
+   *
+   * @param rawText text to be parsed
+   * @return text in a valid messageML format
+   */
+  public static String escapeSpecialChars(String rawText) {
+    Matcher matcher = PATTERN.matcher(rawText);
+    StringBuffer parsedText = new StringBuffer();
```
symphony-youri, line 32: we can use a StringBuilder as we don't need synchronization here

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private static String getReplacementOf(String group) {
+    String replacement = tokens.get(group);
+    if (replacement == null) {
+      replacement = tokens.get("\\" + group);
+    }
+    return replacement;
+  }
+
+  public static String escape(String text) {
```
symphony-youri, line 48: as a user should I use escapeSpecialChars or escape? (something we could clarify with comments maybe?)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
symphony-mariacristina: yes I can use the java.util.Base64

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
symphony-mariacristina: sorry I pushed that by mistake, it was part of my tests to use the StringEscapeUtils but it is not working for every character. I will remove this function since it is working only partially.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
symphony-mariacristina: here is StringBuffer instead of StringBuilder because `appendReplacement` and `appendTail` require both a StringBuffer in input

symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/util/MessageUtilTest.java
symphony-mariacristina: If it's ok I would prefer to keep it like this. Since there are not so much tests and also they are really small.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/stream/util/StreamUtil.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Convert the URLSafe encoded stream id to the corresponding original stream id
+   *
+   * @param streamId of the stream to be parsed
+   * @return stream id after conversion
+   *
+   * <pre>{@code
+   *   String streamId = StreamUtil.fromUrlSafeStreamId("XlU3OH9eVMzq-yss7M_xyn___oxwgbtGbQ");
+   * }</pre>
```
symphony-thibault, line 39: Not sure, but I think it should go before `@param` to be valid Javadoc

symphony-mariacristina: It is visible in both cases, what changes is where it is placed. But I guess it makes more sense to have it right after the description and before the params. I will change that.

symphony-elias (approved):
LGTM !

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java
symphony-youri: Ok I did not know :)

symphony-bdk-core/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
 
     api 'org.apiguardian:apiguardian-api'
     implementation 'org.slf4j:slf4j-api'
+    implementation 'commons-codec:commons-codec'
```
symphony-youri, line 44: we can remove it now (as it has been replaced by base64 from jdk)

symphony-youri (approved):

symphony-thibault (approved):
Go for merge 👍

PR #449, opened 2021-02-17
[FINOS] Configure WhiteSource

symphony-thibault:
WhiteSource configuration from https://github.com/finos/project-blueprint


symphony-youri (approved):

maoo (approved):
LGTM!

In the future we may want to improve the `excludes` and `gradle.ignoredScopes` props, to avoid false positives

Issue #452, opened 2021-02-17
Project update following transfer to FINOS


symphony-thibault:
Now the project has been transferred FINOS, we need to update some stuffs in the code related to the previous location. It will include: 
- [x] links in `README.md`
- [x] Maven coordinates (from `com.symphony.platformsolutions` to `org.finos.symphony.bdk`)
- [x] CICD (from CircleCI/Jenkins to Github actions)

PR #454, opened 2021-02-17
#452 Adding build workflow

symphony-thibault:
### Ticket
Relates to https://github.com/finos/symphony-bdk-java/issues/452

Adding GH Actions in replacement of CircleCI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,8 +19,7 @@ jobs:
     - name: Set up JDK
       uses: actions/setup-java@v1
       with:
-        # 11 runtime needed for Sonar, we still build with 8 as a target
-        java-version: 11
+        java-version: 8
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Build with Gradle
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,11 +16,24 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK
+    - name: Set up JDK 1.8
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v1
+      with:
+        # 11 runtime needed for Sonar, we still build with 8 as a target
+        java-version: 11
+    - name: Grant execute permission for gradlew
```
symphony-youri, line 24: that's weird, issn't it part of git attributes?

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
symphony-youri, line 18: if we want to make it faster we can start to cache dependencies as well

.github/workflows/build.yml
symphony-thibault: I'll try without, I took the template generated with GH itself

.github/workflows/build.yml
symphony-thibault: OK I'll follow https://github.com/symphonyoss/messageml-utils/blob/master/.github/workflows/build.yml#L16

symphony-thibault: 👍  done below

PR #460, opened 2021-02-24
#459 Automated release workflow

symphony-thibault:
### Ticket
https://github.com/finos/symphony-bdk-java/issues/459

### Description
Automated release workflow using Github Actions.


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -0,0 +1,12 @@
+# https://github.com/gradle/wrapper-validation-action#readme
+
+name: "Validate Gradle Wrapper"
```
symphony-youri, line 3: Good question, were we going to accept an external PR with a JAR change?
But then did you check the source code of the action to make sure it was not doing anything suspicious it self 😛

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+# Release any branch to Maven Central.
+# This workflow currently assumes that the target branch is ready to be release (i.e. version is correct)
+
+name: Release
+
+on: [ workflow_dispatch ]
```
symphony-youri, line 6: What about running it once a release is created in GH? So that we don't have to create the release + trigger the job? and triggering the release also means we are more likely to target a branch/tag that is actually a release

symphony-youri: Or maybe we can keep both, upon release and manually

.github/workflows/gradle-wrapper-validation.yml
symphony-thibault: I hope so ! Otherwise we should really contact the Spring Boot team https://github.com/spring-projects/spring-boot/blob/master/.github/workflows/gradle-wrapper-validation.yml !

.github/workflows/release.yml
symphony-thibault: Yep let me try this way!

symphony-thibault: Now done https://github.com/finos/symphony-bdk-java/blob/main/.github/workflows/release.yml#L8

Issue #464, opened 2021-03-02
🔫  Bulletproof DatafeedLoop


symphony-thibault:
### Context
One the most important feature of the BDK is certainly the [Datafeed Loop](https://github.com/finos/symphony-bdk-java/blob/main/docs/datafeed.md) that allows your bot application to receive [Real Time Events](https://developers.symphony.com/restapi/docs/real-time-events). The health of your bot strongly depends on this DF Loop. A broken DF Loop means that your bot cannot receive chat events and therefore cannot properly works. 

### Problem
When using the Spring Boot wrapper, the DF Loop [is executed as a separate thread](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/service/DatafeedAsyncLauncherService.java) in order to not block the main thread. This is quite different for a pure Java application, when the DF Loop is usually started as part of the main thread and blocks this one until the loop (e.g. `while(true) { readDatafeed(); }` is still running). 

We already implemented a dedicated datafeed retry strategy that can be configured as such: 
```yaml
# config.yaml
datafeed:
  retry:
    maxAttempts: 6
    initialIntervalMillis: 2000
    multiplier: 1.5
    maxIntervalMillis: 10000
```
Reaching the `maxAttempts` value means that the loop will exit and thus its thread will stop as well. Causing an unrecoverable state for your bot application. 

❓  We would like to figure out what could be the best strategy to handle this bad situation, when your container is still alive but the DF Loop has crashed making your bot application unusable. 

### Proposed solution
One of the most straightforward solution that came to our mind is to make `maxAttemps` value almost infinite (e.g. equals to `Integer.MAX_VALUE`). Depending on the others retry parameters (`initialIntervalMillis`, `multiplier` and `maxIntervalMillis`) this would make the loop retrying for several years, so let's say forever! 

#### Config issue
Let's imagine that your BDK config contains an issue (for instance a wrong pod url). Your container successfully starts up but the DF Loop directly starts retrying forever, making your container-orchestrator unable to detect any issue.

symphony-youri:
I would perhaps go with maxAttemps=-1 as the default for the DF loop in order to express infinite retries

PR #469, opened 2021-03-07
#464 Infinite number of attempts for the Datafeed retry configuration

symphony-thibault:
### Ticket
Closes https://github.com/finos/symphony-bdk-java/issues/464

### Description
Now datafeed is by default configured to have an infinite number of retry attempts. Global retry configuration keeps the same default value (`10`). 

### Checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -119,7 +119,7 @@ datafeed:
     maxIntervalMillis: 10000
 
 retry:
-  maxAttempts: 6
+  maxAttempts: 6 # set '-1' for an infinite number of attempts, default value is '10'
   initialIntervalMillis: 2000
   multiplier: 1.5
   maxIntervalMillis: 10000
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRetryConfig.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRetryConfig.java
@@ -1,28 +1,66 @@
 package com.symphony.bdk.core.config.model;
 
-import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
 import org.apiguardian.api.API;
 
-@Getter
 @Setter
+@NoArgsConstructor
 @API(status = API.Status.STABLE)
 public class BdkRetryConfig {
 
-    public static final int DEFAULT_MAX_ATTEMPTS = 10;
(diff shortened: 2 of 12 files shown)


symphony-youri:
Is it expected that we don't a check building the project?

docs/configuration.md
```diff
@@ -119,7 +119,7 @@ datafeed:
     maxIntervalMillis: 10000
 
 retry:
-  maxAttempts: 6
+  maxAttempts: 6 # set '-1' for an infinite number of attempts, default value is '10'
```
symphony-youri, line 122: default is infinite isn't it?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRetryConfig.java
```diff
@@ -1,28 +1,66 @@
...
+
+  public Integer getMaxAttempts() {
+
+    if (this.maxAttempts == null) {
+      return DEFAULT_MAX_ATTEMPTS;
+    }
+
+    if (this.maxAttempts < 0) {
+      // negative value means infinite number of attempts
+      return Integer.MAX_VALUE;
```
symphony-youri, line 35: Does it last until the sun dies? (well unless Elon gets us out of the solar system)

symphony-youri (approved):

docs/configuration.md
symphony-thibault: For datafeed retry default is `-1`, otherwise for global retry we kept `10`. Is it something we should align? I was just thinking that an infinite retry is not something that we would expect when calling regular endpoints

symphony-thibault:
> Is it expected that we don't a check building the project?

Just fixed https://github.com/finos/symphony-bdk-java/pull/469/commits/1bfafa3ef93c8a6482d73a5913c2646d86eb0e43

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkRetryConfig.java
symphony-thibault: Haha, with 2s between each attempt that would give us 8171 years to fix a pod failure :)

symphony-thibault:
@symphony-youri I'd maybe backport it to `2.1.x` branch and then release `2.1.2`. WDYT?

docs/configuration.md
symphony-youri: my bad I thought it was for the DF loop only

PR #470, opened 2021-03-08
#464 Infinite number of attempts for the Datafeed retry configuration

symphony-thibault:
Backported from https://github.com/finos/symphony-bdk-java/pull/469


symphony-elias (approved):
LGTM

PR #480, opened 2021-03-30
PLAT-10720 - Add configuration to disable reusing datafeed id

anthony-symphony:
### Ticket
PLAT-10720

### Description
Add configuration value to bypass the reusing of datafeed id as some sites do not handle this properly.

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
@@ -11,17 +11,26 @@ import java.io.File;
 @API(status = API.Status.STABLE)
 public class BdkDatafeedConfig {
 
-    private String version = "v1";
-    private String idFilePath;
-    private BdkRetryConfig retry = new BdkRetryConfig(BdkRetryConfig.INFINITE_MAX_ATTEMPTS);
+  private String version = "v1";
+  private String idFilePath;
+  private BdkRetryConfig retry = new BdkRetryConfig(BdkRetryConfig.INFINITE_MAX_ATTEMPTS);
+  private Boolean reuseDatafeedId;
 
-    public String getIdFilePath() {
-        if (idFilePath == null || idFilePath.isEmpty()) {
-            return "." + File.separator;
-        }
-        if (!idFilePath.endsWith(File.separator)) {
-            return idFilePath + File.separator;
-        }
-        return idFilePath;
+  public String getIdFilePath() {
+    if (idFilePath == null || idFilePath.isEmpty()) {
+      return "." + File.separator;
     }
+    if (!idFilePath.endsWith(File.separator)) {
+      return idFilePath + File.separator;
+    }
+    return idFilePath;
+  }
+
+  public boolean getReuseDatafeedId() {
+    if (reuseDatafeedId == null) {
+      return true;
+    } else {
+      return reuseDatafeedId;
+    }
+  }
 }
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
@@ -1,0 +1,58 @@
+package com.symphony.bdk.core.service.datafeed.impl;
+
+import com.symphony.bdk.core.config.model.BdkConfig;
+import com.symphony.bdk.core.service.datafeed.DatafeedIdRepository;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.apiguardian.api.API;
+
+/**
+ * The implementation of {@link DatafeedIdRepository} interface for a
+ * non-persistent in memory Datafeed.
+ */
+@Slf4j
+@API(status = API.Status.INTERNAL)
+public class InMemoryDatafeedIdRepository implements DatafeedIdRepository {
+
+  private final BdkConfig config;
+  private String datafeedId = null;
+  private String agentBasePath = null;
+
+  public InMemoryDatafeedIdRepository(BdkConfig config) {
+    this.config = config;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void write(String datafeedId) {
+    write(datafeedId, this.config.getAgent().getBasePath());
+  }
+
+  public void write(String datafeedId, String agentBasePath) {
+    log.debug("Writing datafeed id {} to memory.", datafeedId);
+    this.datafeedId = datafeedId;
+    this.agentBasePath = agentBasePath;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Optional<String> read() {
+    if (datafeedId == null) {
+      return Optional.empty();
+    } else {
+      return Optional.of(datafeedId);
+    }
+  }
+
+  public Optional<String> readAgentBasePath() {
+    if (agentBasePath == null) {
+      return Optional.empty();
+    } else {
+      return Optional.of(agentBasePath);
+    }
+  }
+}
--- a/symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigTest.java
+++ b/symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigTest.java
@@ -1,9 +1,7 @@
 package com.symphony.bdk.core.config;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
 
 import com.symphony.bdk.core.client.exception.ApiClientInitializationException;
 import com.symphony.bdk.core.config.model.BdkBotConfig;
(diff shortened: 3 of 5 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Optional<String> read() {
+    if (datafeedId == null) {
+      return Optional.empty();
+    } else {
+      return Optional.of(datafeedId);
+    }
```
symphony-thibault, line 48: `return Optional.ofNullable(this.datafeedId)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public Optional<String> readAgentBasePath() {
+    if (agentBasePath == null) {
+      return Optional.empty();
+    } else {
+      return Optional.of(agentBasePath);
+    }
```
symphony-thibault, line 56: `return Optional.ofNullable(this. agentBasePath)`

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void write(String datafeedId) {
+    write(datafeedId, this.config.getAgent().getBasePath());
+  }
+
+  public void write(String datafeedId, String agentBasePath) {
```
symphony-thibault, line 33: `@Override` is missing right?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apiguardian.api.API;
+
+/**
+ * The implementation of {@link DatafeedIdRepository} interface for a
+ * non-persistent in memory Datafeed.
+ */
+@Slf4j
+@API(status = API.Status.INTERNAL)
+public class InMemoryDatafeedIdRepository implements DatafeedIdRepository {
```
symphony-thibault, line 15: Could you then remove the one from [test sources](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/test/java/com/symphony/bdk/core/test/InMemoryDatafeedIdRepository.java)? I guess that it will require to update unit tests as well

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
```diff
@@ -11,17 +11,26 @@
...
+    if (idFilePath == null || idFilePath.isEmpty()) {
+      return "." + File.separator;
     }
+    if (!idFilePath.endsWith(File.separator)) {
+      return idFilePath + File.separator;
+    }
+    return idFilePath;
+  }
+
+  public boolean getReuseDatafeedId() {
```
symphony-thibault, line 29: I'd perhaps rename it to `isReuseDatafeedId()`, WDYT?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
```diff
@@ -11,17 +11,26 @@
 @API(status = API.Status.STABLE)
 public class BdkDatafeedConfig {
 
-    private String version = "v1";
-    private String idFilePath;
-    private BdkRetryConfig retry = new BdkRetryConfig(BdkRetryConfig.INFINITE_MAX_ATTEMPTS);
+  private String version = "v1";
+  private String idFilePath;
+  private BdkRetryConfig retry = new BdkRetryConfig(BdkRetryConfig.INFINITE_MAX_ATTEMPTS);
+  private Boolean reuseDatafeedId;
```
symphony-thibault, line 17: You will also need to reflect this change in https://github.com/finos/symphony-bdk-java/blob/main/docs/configuration.md#full-configuration-example

symphony-thibault (changes requested):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
anthony-symphony: Followed https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java so we need to fix it there as well if we need to.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
anthony-symphony: Followed https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java so we need to fix it there as well if we need to.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
anthony-symphony: Followed https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/OnDiskDatafeedIdRepository.java so we need to fix it there as well if we need to.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
anthony-symphony: Sure.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/InMemoryDatafeedIdRepository.java
anthony-symphony: Not sure what you mean here

anthony-symphony: Oh I didn't even notice that was there.

docs/configuration.md
```diff
@@ -180,6 +181,7 @@ configuration inside service configuration to override the global one.
 The datafeed configuration will contain information about the datafeed service to be used by the bot:
 - `version`: the version of datafeed service to be used. By default, the bot will use the datafeed v1
 service. 
+- `reuseDatafeedId`: specifies if we should read/write from `idFilePath`. If not specified, this is defaulted to true.
```
symphony-thibault, line 184: + (`available for Datafeed v1 only`)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
```diff
@@ -11,17 +12,28 @@
...
+    if (!idFilePath.endsWith(File.separator)) {
+      return idFilePath + File.separator;
+    }
+    return idFilePath;
+  }
+
+  public boolean isReuseDatafeedId() {
+    if (reuseDatafeedId == null) {
+      return true;
+    } else {
```
symphonydarlys, line 35: we do not need "else" here we can return directly reuseDatafeedId.

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigTest.java
```diff
@@ -98,4 +96,16 @@ void checkDefaultDatafeedRetryConfiguration() throws Exception {
...
   }
+
+  @Test
+  void checkDatafeedPersistenceFalse() throws Exception {
+    final BdkConfig config = BdkConfigLoader.loadFromClasspath("/config/df_no_reuse.yaml");
+    assertFalse(config.getDatafeed().isReuseDatafeedId());
+  }
+
+  @Test
+  void checkDatafeedPersistenceNotSet() throws Exception {
```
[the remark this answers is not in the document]
anthony-symphony: I have one right above where it is set to false? You want another one to check if we set true it detects it as true?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
symphony-thibault: Well, as it is declared as `private Boolean reuseDatafeedId`, we still need this check to handle the case when the field hasn't been set in configuration. 

Maybe it would worth checking what happen if we declare `private boolean reuseDatafeedId = true;` (_including the default backward compatible value_). 

Otherwise I'd still have a preference for `return this.reuseDatafeedId == null ? true : this.reuseDatafeedId` :)

anthony-symphony: `private boolean reuseDatafeedId = true;` works actually so I am going with that.

symphony-youri:
Hello @anthony-symphony, has the need for such feature in DFv1 come up again?

PR #486, opened 2021-04-02
#483 Jwt Cookie set with SameSite=Strict

### LINKED ISSUE #483: Cookie set without SameSite=Strict
## Bug Report

The request to 
**/bdk/v1/app/auth**
 appears to be vulnerable to cross-site request forgery (CSRF) attacks against authenticated users.

The request can be issued cross-domain.
The BDK relies solely on HTTP cookies to identify the user that issued the request.
The request performs some privileged action within the application, which returns a token.
The attacker can determine all the parameters required to construct a request that performs the action. If the request contains any values that the attacker cannot determine or predict, then it is not vulnerable.

### Steps to Reproduce:
1. .Send an HTTP request to **/bdk/v1/app/auth** by modifying the Referer (see below)
image
2. The request passes and the token is obtained:
image

### Expected Result:
Should return a 401 (Not authenticated), because the Cookie should not be sent (auth cookie not sent = not authenticated)

### Actual Result:
The request passes through, because of the Cookie which is sent, due to the fact that it does not contain the Strict value for Samesite attribute (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict).

### Environment:
None

### Additional Context:
None

### Recommendation:
Set the Strict value to the Samesite attribute. This will prevent sending the cookie when the request is done from another domain.

symphony-hong:
### Ticket
Closes #483

### Description
The request to /bdk/v1/app/auth appears to be vulnerable
to cross-site request forgery (CSRF) attacks against authenticated users.

The request can be issued cross-domain.
The BDK relies solely on HTTP cookies to identify the user that issued the request.
The request performs some privileged action within the application, which returns a token.
The attacker can determine all the parameters required to construct a request that
performs the action. If the request contains any values that the attacker cannot
determine or predict, then it is not vulnerable.

Setting the SameSite=Strict will prevent sending the cookie when the request is done from another domain.

### Dependencies

### Checklist

### DIFF
--- a/docs/spring-boot/app-starter.md
+++ b/docs/spring-boot/app-starter.md
@@ -91,6 +91,7 @@ bdk-app:
       enabled: true # activate the CircleOfTrust endpoints (default is true)
       jwtCookie:
         enabled: true # activate the jwt cookie storage (default is false)
+        sameSite: Strict # same site configuration to restrict the jwt cookie from cross-site domain (default is Strict)
         expireIn: 1d # jwt cookie duration (default value is 1d, see https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration) 
     cors: # enable Cross-Origin Resource Sharing (CORS) communication
       "[/**]": # url mapping
--- a/symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
+++ b/symphony-bdk-examples/bdk-app-spring-boot-example/src/main/resources/application.yaml
@@ -18,6 +18,7 @@ bdk-app:
     jwtCookie:
       enabled: true
       maxAge: 1d
+      sameSite: None
   cors:
     "[/**]":
       allowed-origins: "*"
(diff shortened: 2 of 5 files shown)


symphony-thibault (approved):
LGTM !

symphony-youri:
Isn't it going to break https://perzoinc.atlassian.net/browse/APP-3271 ?

symphony-hong:
IMHO, what we did in the APP-3271 is not the right way to perform the cross-site call in security sense. 
Maybe, we can change the SameSite default to Strict and make it configurable for someone who want to do it without caring about the security issue?

symphony-elias (approved):
LGTM

symphony-youri:
Does someone remember why we set the userJwt cookie in the first place by the way?

symphony-thibault (approved):
Thanks for this contribution :)

symphony-hong:
> Does someone remember why we set the userJwt cookie in the first place by the way?

It is the existing behaviour from the legacy bdk when we developed COT for BDK 2.0, I'm not sure about the root reason.

symphony-thibault:
> Does someone remember why we set the userJwt cookie in the first place by the way?

For "backward compatibility" I'd say. User session could also be stored in local storage though, however cookies are stored automatically and therefore doesn't require any additional code on client side.

symphony-thibault:
@symphony-hong, sorry I forgot to tell you, but could you backport this change to `2.1-rc` branch?

PR #490, opened 2021-04-15
APP-3296 PLAT-10766 Add .devx-cicd/config.yaml file, that builds snapshot artifacts (#370)

micaste:
For now, this devx-cicd tasks only builds and run unit-tests as a check. But it should later include e2e tests as well


micaste:
Closing in favor of a private repository for internal automation in the future

PR #492, opened 2021-04-16
Run multiple instances of a bot

### LINKED ISSUE #488: Support running multiple instances of a bot
### Description of Problem:
For availability and scalability it would be useful to be able to run multiple instances of a single bot.

The main problem lies in the handling of Datafeed events, we want to make sure that only one instance gets an event and if it fails to process it that this event gets dispatched to another instance.

### Potential Solutions:

Datafeed v2 API exposes a tag parameter that would allow a bot to create a single instance of a datafeed and the events would be randomly dispatched to any instance reading events from this feed.

symphony-youri:
Closes #488 

### Description
By supporting the tag parameter of the DFv2 API we can support multiple instances of a bot, each one receiving the events in turn.

One of the impacting change here is in the DF loop where the ack id is only updated after all events have been processed.
This way the events are re-queued.
**Do we want that?**

It means in case a listener fails, all events from a read will be requeued with a risk of processing them twice.
Should we make it explicit for the bot's developer? with a specific exception to throw in the case you want to reprocess the event elsewhere?

The tag is taken from the bot's username, there is a limit on 100 characters but IMHO we rarely go over that limit.
If it happens an admin would have to make sure the first 100 characters of the username are unique.
If this is still an issue then we might consider letting the bot developer set this tag and default to the username.

The examples are provided part of the multi-instances example module.

### Dependencies
This will work once https://github.com/symphonyoss/symphony-api-spec/pull/112 is merged.
Hence the specs URL has been changed to use the master branch.

### Checklist
- [X] Filled properly the description and dependencies, if any
- [X] Unit tests updated or added
- [X] Javadoc added or updated
- [X] Updated the documentation in [docs folder](../docs)


symphony-thibault:
Could we provide an optional (local) retention cache of already processed messages? In order to avoid listeners to process a message twice.

symphony-bdk-examples/bdk-multi-instances-example/build.gradle
```diff
@@ -12,6 +12,8 @@ dependencies {
     implementation 'org.apache.commons:commons-lang3'
     implementation 'org.springframework.boot:spring-boot-starter'
 
+    implementation 'com.hazelcast:hazelcast:4.2'
+
     implementation 'org.springframework.kafka:spring-kafka:2.6.5'
```
symphony-thibault, line 17: Let's remove kafka implementation then!

symphony-youri:
> Could we provide an optional (local) retention cache of already processed messages? In order to avoid listeners to process a message twice.

If we don't want to implement this cache ourselves that would mean to add a dependency like caffeine (to make the map bounded in size so it does not grow infinitely)

symphony-youri:
Build will fail until the spec is updated

symphony-bdk-examples/bdk-multi-instances-example/build.gradle
symphony-youri: Done

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedLoopV2.java
```diff
@@ -45,13 +47,24 @@
...
   private final AtomicBoolean started = new AtomicBoolean();
   private final AckId ackId;
+  private final String tag;
+
   private V5Datafeed datafeed;
 
   public DatafeedLoopV2(DatafeedApi datafeedApi, AuthSession authSession, BdkConfig config) {
     super(datafeedApi, authSession, config);
     this.ackId = new AckId().ackId("");
+    this.tag = StringUtils.truncate(bdkConfig.getBot().getUsername(), DATAFEED_TAG_MAX_LENGTH);
```
symphony-elias, line 67: Should we log a warning if the bot username is greater than ` DATAFEED_TAG_MAX_LENGTH` ?

symphony-elias (approved):
Looks good

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedLoopV2.java
symphony-youri: We could but at the same time it could be a valid use case if you made sure that your bot's name starts with a differentiating part (bot1_longstring... and bot2_longstring...)

What do you think @symphony-thibault ?

symphony-thibault: As `bdkConfig.getBot().getUsername()` is the value used for authentication, I don't really see how we could end up with `bot1_longstring...` and `bot2_longstring...`

symphony-youri: Something else to consider, the tag is specific for each user. So it is very unlikely that truncating the bot's service account name to <100 characters is going to be a problem as most of the time a service account is used for a single bot.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedLoopV1.java
```diff
@@ -114,7 +114,11 @@ private Void readAndHandleEvents() throws ApiException {
         datafeedApi.v4DatafeedIdReadGet(datafeedId, authSession.getSessionToken(), authSession.getKeyManagerToken(),
             null);
     if (events != null && !events.isEmpty()) {
-      handleV4EventList(events);
+      try {
+        handleV4EventList(events);
+      } catch (RequeueEventException e) {
+        log.warn("EventException is not supported for DFv1, events will not get re-queued", e);
```
symphony-thibault, line 120: Is it something we want to log at `warn` level ?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityMatcher.java
```diff
@@ -15,7 +17,7 @@
    * @param context Current activity context.
    * @return true if {@link AbstractActivity#onActivity(ActivityContext)} can be triggered, false otherwise.
```
symphony-thibault, line 18: I'd add the `@throws` Javadoc here as well

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/RequeueEventException.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.apiguardian.api.API;
+
+/**
+ * Internal exception to convey the message that ack id should not be updated thus re-queuing events.
+ */
+@API(status = API.Status.INTERNAL)
+public class RequeueEventException extends RuntimeException {
+  public RequeueEventException(V4Event event, RealTimeEventListener listener, EventException e) {
+    super(String.format("Listener %s failed to process event %s, events will be re-queued",
```
symphony-thibault, line 15: Let's log the event type as well, that would help for debugging

symphony-youri: +1

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/impl/DatafeedLoopV1.java
symphony-youri: I would say yes because it won't prevent the bot to work (or from the bdk point of view there is nothing that can be done to recover from that) but might lead to unexpected behavior

symphony-youri: or were you thinking about lowering the level to info?

symphony-bdk-examples/bdk-app-spring-boot-example/build.gradle
```diff
@@ -10,5 +10,3 @@ dependencies {
     implementation project(':symphony-bdk-spring:symphony-bdk-app-spring-boot-starter')
     implementation 'org.springframework.boot:spring-boot-starter-security'
 }
-
-uploadArchives.enabled = false
```
symphony-elias, line 14: Why did you remove that ? Is it already disabled in the root build.gradle file ?

symphony-youri: This is coming from an old plugin that we don't use anymore (and now this API has been removed in Gradle 7)
Now it works with convention plugins where we explictly apply the id 'bdk.java-publish-conventions' plugin
So we don't have to disable the upload manually

symphony-thibault (approved):
Great work !

Issue #506, opened 2021-04-26
Allow environment variables in config file


Yannick-Symphony:
## Feature Request
Allow environment variables in config file

### Description of Problem:
One common practice when deploying applications with docker is to define deployment-specific configuration as environment variables
This works fine for standard spring applications, as spring will read ${env_var} in config files. however it does not work with the current jackson parse of the BDK

### Potential Solutions:
Modify the existing config parser to try to match any config key that hasn't already matched a java variable ( https://github.com/finos/symphony-bdk-java/pull/403 ) with environment variables

PR #507, opened 2021-04-26
Allow environment variables in config file

Yannick-Symphony:
### Ticket
fixes https://github.com/finos/symphony-bdk-java/issues/506

### Description
Allow environment variables in config files

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -226,10 +226,11 @@ look like:
 }
 ``` 
 
-### Field interpolation using system properties
-In both formats, you can use system properties as field values. For instance, `${user.home}` in any field will be
-replaced by the actual value of system property `user.home`. If a property is not defined, no interpolation will be done
-and string will be left as is. A default value can be provided after `:-`, for instance `${property.name:-defaultValue}`.
+### Field interpolation using java system properties or environment variables
+In both formats, you can use java system properties and system environment variables as field values. For instance, `${user.home}` in any field will be
+replaced by the actual value of java system property `user.home`. Likewise for `$HOME`, mapping the environment variable `HOME`. If a property is not defined, no interpolation will be done
+and string will be left as is. If the same key is defined both as a java system property and an environment variable, the java property value will take precedence.
+A default value can be provided after `:-`, for instance `${property.name:-defaultValue}`.
 Therefore, the following is a valid configuration file:
 
 ```json
@@ -238,12 +239,14 @@ Therefore, the following is a valid configuration file:
   "bot": {
     "username": "bot-username",
     "privateKey": {
-      "path": "${user.home}/rsa-private-key.pem"
+      "path": "${HOME}/rsa-private-key.pem"
     }
   }
 }
 ```
 Please mind that if you want to escape the `$` sign, `$${value}` will be replaced by `${value}`.
+And as the matching of environment variables is done after java system properties, if you have a system property with **value** `${value}` and en environment variable with **key** `value`, it will substitute the value of the environment variable
+
 
 Reading a `JSON` configuration file is completely transparent: 
 ```java
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -226,10 +226,10 @@ look like:
 }
 ``` 
 
-### Field interpolation using java system properties or environment variables
-In both formats, you can use java system properties and system environment variables as field values. For instance, `${user.home}` in any field will be
-replaced by the actual value of java system property `user.home`. Likewise for `$HOME`, mapping the environment variable `HOME`. If a property is not defined, no interpolation will be done
-and string will be left as is. If the same key is defined both as a java system property and an environment variable, the java property value will take precedence.
+### Field interpolation using Java system properties or environment variables
+In both formats, you can use Java system properties and system environment variables as field values. For instance, `${user.home}` in any field will be
+replaced by the actual value of Java system property `user.home`. Likewise for `$HOME`, mapping the environment variable `HOME`. If a property is not defined, no interpolation will be done
+and string will be left as is. If the same key is defined both as a Java system property and an environment variable, the Java property value will take precedence.
 A default value can be provided after `:-`, for instance `${property.name:-defaultValue}`.
 Therefore, the following is a valid configuration file:
 
@@ -245,7 +245,7 @@ Therefore, the following is a valid configuration file:
 }
 ```
 Please mind that if you want to escape the `$` sign, `$${value}` will be replaced by `${value}`.
-And as the matching of environment variables is done after java system properties, if you have a system property with **value** `${value}` and en environment variable with **key** `value`, it will substitute the value of the environment variable
+And as the matching of environment variables is done after Java system properties, if you have a system property with **value** `${value}` and en environment variable with **key** `value`, it will substitute the value of the environment variable
 
 
 Reading a `JSON` configuration file is completely transparent: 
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -226,10 +226,11 @@ look like:
 }
 ``` 
 
-### Field interpolation using system properties
-In both formats, you can use system properties as field values. For instance, `${user.home}` in any field will be
-replaced by the actual value of system property `user.home`. If a property is not defined, no interpolation will be done
-and string will be left as is. A default value can be provided after `:-`, for instance `${property.name:-defaultValue}`.
+### Field interpolation using java system properties or environment variables
+In both formats, you can use java system properties and system environment variables as field values. For instance, `${user.home}` in any field will be
+replaced by the actual value of java system property `user.home`. Likewise for `$HOME`, mapping the environment variable `HOME`. If a property is not defined, no interpolation will be done
```
symphony-youri, line 231:
```suggestion
replaced by the actual value of Java system property `user.home`. Likewise for `$HOME`, mapping the environment variable `HOME`. If a property is not defined, no interpolation will be done
```

docs/configuration.md
```diff
@@ -226,10 +226,11 @@ look like:
...
+and string will be left as is. If the same key is defined both as a java system property and an environment variable, the java property value will take precedence.
```
symphony-youri, line 232:
```suggestion
and string will be left as is. If the same key is defined both as a Java system property and an environment variable, the Java property value will take precedence.
```

docs/configuration.md
```diff
@@ -238,12 +239,14 @@ Therefore, the following is a valid configuration file:
...
     "username": "bot-username",
     "privateKey": {
-      "path": "${user.home}/rsa-private-key.pem"
+      "path": "${HOME}/rsa-private-key.pem"
     }
   }
 }
 ```
 Please mind that if you want to escape the `$` sign, `$${value}` will be replaced by `${value}`.
+And as the matching of environment variables is done after java system properties, if you have a system property with **value** `${value}` and en environment variable with **key** `value`, it will substitute the value of the environment variable
```
symphony-youri, line 248:
```suggestion
And as the matching of environment variables is done after Java system properties, if you have a system property with **value** `${value}` and en environment variable with **key** `value`, it will substitute the value of the environment variable
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
```diff
@@ -27,10 +28,12 @@
 
     private static final ObjectMapper JSON_MAPPER = new JsonMapper();
     private static final ObjectMapper YAML_MAPPER = new YAMLMapper();
+    private static final StringSubstitutor envVarStringSubstitutor;
 
     static {
         JSON_MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
         YAML_MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        envVarStringSubstitutor = new StringSubstitutor(StringLookupFactory.INSTANCE.environmentVariableStringLookup());
```
symphony-youri, line 36: In the doc they mention that StringSubstitutor is not thread safe, could it be a problem?

maybe to be safe I would change BdkConfigParser to create an instance when parsing a config instead of having it as a static utils

symphony-youri (commented):
Thanks for the PR!

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigParserTest.java
```diff
@@ -11,13 +11,34 @@
...
+    void hackEnvVar(String key, String value) throws NoSuchFieldException, IllegalAccessException {
+        Map<String, String> env = System.getenv();
+        Field field = env.getClass().getDeclaredField("m");
+        field.setAccessible(true);
+        if(value==null) {
+          ((Map<String, String>) field.get(env)).remove(key);
+        }
+        else {
+          ((Map<String, String>) field.get(env)).put(key, value);
+        }
```
symphony-thibault, line 29: Could you reformat the file, our [`.editorconfig`](https://github.com/finos/symphony-bdk-java/blob/main/.editorconfig) should give: 
```java
    Map<String, String> env = System.getenv();
    Field field = env.getClass().getDeclaredField("m");
    field.setAccessible(true);
    if (value == null) {
      ((Map<String, String>) field.get(env)).remove(key);
    } else {
      ((Map<String, String>) field.get(env)).put(key, value);
    }
```

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
symphony-elias: I found that: https://stackoverflow.com/a/2412144

symphony-youri: the variable will be thread safe in the sense it will be created only once but the usage of StringSubstitutor by multiple threads won't

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigParserTest.java
```diff
@@ -11,112 +11,166 @@
...
-        String configPath = "/config/invalid_config.yaml";
-        BdkConfigException exception = assertThrows(BdkConfigException.class, () -> {
-            InputStream inputStream = BdkConfigLoaderTest.class.getResourceAsStream(configPath);
-            BdkConfigParser.parse(inputStream);
-        });
-        assertEquals("Given InputStream is not valid. Only YAML or JSON are allowed.", exception.getMessage());
+  }
+
+  @BeforeEach
+  void tearDown() {
```
symphony-elias, line 32: `tearDown` for a `@BeforeEach` method is a bit misleading IMHO. `setUp` is a more usual name for that kind of method.

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigParserTest.java
```diff
@@ -11,112 +11,166 @@
...
+    System.setProperty("my.property", "propvalue");
+    hackEnvVar("SYMPHONY_BDK_UNIT_TEST_TEMP", "envvalue");
+    //java vars should take precedence over env vars
+    hackEnvVar("my.property", "invalid_value");
+    String configPath = "/config/config_properties.yaml";
+    InputStream inputStream = BdkConfigLoaderTest.class.getResourceAsStream(configPath);
+    BdkConfigParser parser = new BdkConfigParser();
+    JsonNode jsonNode = parser.parse(inputStream);
+    assertEquals("envvalue/propvalue/privatekey.pem", jsonNode.at("/bot/privateKey/path").asText());
+    ;
```
symphony-elias, line 120: Please remove this semicolon

symphony-bdk-core/src/test/java/com/symphony/bdk/core/config/BdkConfigParserTest.java
Yannick-Symphony: I agree!

https://github.com/finos/symphony-bdk-java/commit/458eb2687ee4186651aba8093ee7016884ae2a4d#diff-770895de718032b550984a565ad529dae57cb73e6d7288826459ee1c09248528

symphony-bdk-core/src/main/java/com/symphony/bdk/core/config/BdkConfigParser.java
Yannick-Symphony: ok I moved it to instance-based system

Yannick-Symphony:
updated after second round with @symphony-youri and @symphony-elias

symphony-youri (approved):

Issue #504, opened 2021-04-21
HTTP 500 errors should be retried


symphony-elias:
## Bug Report
During maintenance windows, when pod, km or agent is down, HTTP calls can result in HTTP 500. Such calls need to be retried.
For instance, on one of our bots, we got:
```
2021-04-20 18:26:26.546  INFO 1 --- [_DatafeedThread] c.s.b.c.s.datafeed.impl.DatafeedLoopV1   : Recreate a new datafeed and try again
2021-04-20 18:26:26.684  INFO 1 --- [_DatafeedThread] .s.b.c.r.r.Resilience4jRetryWithRecovery : Retry in 64.0s...
2021-04-20 22:01:45.235  INFO 1 --- [_DatafeedThread] c.s.b.c.s.datafeed.impl.DatafeedLoopV1   : Recreate a new datafeed and try again
2021-04-20 22:01:45.389 ERROR 1 --- [_DatafeedThread] c.s.b.s.s.DatafeedAsyncLauncherService   : An API error has been received while starting the Datafeed loop in a separate thread, please check error below:

com.symphony.bdk.http.api.ApiException: {"code":500,"message":"Received an error when calling a pod endpoint"}
        at com.symphony.bdk.http.jersey2.ApiClientJersey2.invokeAPI(ApiClientJersey2.java:168) ~[symphony-bdk-http-jersey2-2.1.3.jar:2.1.3]
        at com.symphony.bdk.gen.api.DatafeedApi.v4DatafeedCreatePostWithHttpInfo(DatafeedApi.java:479) ~[symphony-bdk-core-2.1.3.jar:2.1.3]
        at com.symphony.bdk.gen.api.DatafeedApi.v4DatafeedCreatePost(DatafeedApi.java:414) ~[symphony-bdk-core-2.1.3.jar:2.1.3]
        at com.symphony.bdk.core.service.datafeed.impl.DatafeedLoopV1.createDatafeedAndPersist(DatafeedLoopV1.java:144) ~[symphony-bdk-core-2.1.3.jar:2.1.3]
        at com.symphony.bdk.core.retry.RetryWithRecovery.executeOnce(RetryWithRecovery.java:105) ~[symphony-bdk-core-2.1.3.jar:2.1.3]
        at io.github.resilience4j.retry.Retry.lambda$decorateCheckedSupplier$3f69f149$1(Retry.java:137) ~[resilience4j-retry-1.6.1.jar:1.6.1]
[...]

PR #509, opened 2021-04-29
#504 DatafeedLoop now retries on 500 server errors

symphony-thibault:
### Ticket
Fixes https://github.com/finos/symphony-bdk-java/issues/504

### Description
We recently noticed that the DatafeedLoop was crashing with `500` errors returned from Agent. This is something we initially stated that we did not wanted to retry on this specific status. Now we do :) It has actually been raised by some customers. 

Will have to be backported to `2.1-rc` branch.


symphonydarlys (approved):

symphony-youri (commented):
Should we add test to check that 500 is indeed a server error?
(also in the commit it would be interesting to describe why we made the change, when I look at Git's history I don't want to navigate to a PR :))

symphony-youri (approved):

symphony-elias (approved):

Issue #513, opened 2021-05-14
Log ssl certs loaded from trustStore


symphony-soufiane:
## Feature Request
Log ssl certs that have been successfully loaded by the bot.

### Description of Problem:
We would like to be able to check if the certs have been loaded by the bot successfully. The feature that the certs loaded from trustStore are printed out into the bot logs was added to the old Java SDK since version 1.1.5 https://github.com/finos/symphony-bdk-java/releases/tag/1.1.5, but this is missing in BDK2.0.

PR #514, opened 2021-05-14
#513 log ssl certs loaded by the bot

symphony-soufiane:
### Ticket
[#513 ](https://github.com/finos/symphony-bdk-java/issues/513)

### Description
In order to help developers to debug ssl certs issues, a log has been added when the certs are loaded from the trustStore.

### DIFF
--- a/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
+++ b/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
@@ -19,21 +19,15 @@ import org.glassfish.jersey.media.multipart.MultiPartFeature;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 import javax.net.ssl.SSLContext;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Specific implementation of {@link ApiClientBuilder} which creates a new instance of an {@link ApiClientJersey2}.
@@ -45,7 +39,6 @@ import org.slf4j.LoggerFactory;
 @API(status = API.Status.STABLE)
 public class ApiClientBuilderJersey2 implements ApiClientBuilder {
 
-  private static final Logger logger = LoggerFactory.getLogger(ApiClientBuilderJersey2.class);
   private static final String TRUSTSTORE_FORMAT = "JKS";
 
   protected String basePath;
@@ -264,30 +257,15 @@ public class ApiClientBuilderJersey2 implements ApiClientBuilder {
     try {
       SSLContext sslContext = sslConfig.createSSLContext();
 
-      // if logging debug is enabled, we print the truststore entries
-      if (logger.isDebugEnabled()) {
-        this.logTrustStore();
+      if (isNotEmpty(trustStoreBytes) && isNotEmpty(trustStorePassword)) {
+        ApiUtils.createAndLogTrustStore(TRUSTSTORE_FORMAT, new ByteArrayInputStream(trustStoreBytes),
+            trustStorePassword.toCharArray());
       }
 
       return sslContext;
-    } catch (IllegalStateException e) {
+    } catch (IllegalStateException | KeyStoreException | IOException | NoSuchAlgorithmException | CertificateException e) {
       throw new IllegalStateException(e.getCause().getMessage(), e);
     }
   }
 
-  private void logTrustStore() {
-    try {
-      if (isNotEmpty(trustStoreBytes) && isNotEmpty(trustStorePassword)) {
-        final KeyStore truststore = KeyStore.getInstance(TRUSTSTORE_FORMAT);
-        truststore.load(new ByteArrayInputStream(trustStoreBytes), trustStorePassword.toCharArray());
-        final List<String> aliases = Collections.list(truststore.aliases());
-        logger.debug("Your custom truststore contains {} entries :", aliases.size());
-        for (String alias : aliases) {
-          logger.debug("# {}", alias);
-        }
-      }
-    } catch (KeyStoreException | NoSuchAlgorithmException | CertificateException | IOException e) {
-      throw new RuntimeException(e);
-    }
-  }
 }
--- a/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
+++ b/symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
@@ -254,11 +254,6 @@ public class ApiClientBuilderJersey2 implements ApiClientBuilder {
       sslConfig
           .trustStoreBytes(trustStoreBytes)
           .trustStorePassword(trustStorePassword);
-
-      // if logging debug is enabled, we print the truststore entries
-      if (logger.isDebugEnabled()) {
-        this.logTrustStore();
-      }
     }
 
     if (isNotEmpty(keyStoreBytes) && isNotEmpty(keyStorePassword)) {
@@ -267,7 +262,14 @@ public class ApiClientBuilderJersey2 implements ApiClientBuilder {
           .keyStorePassword(keyStorePassword);
     }
     try {
-      return sslConfig.createSSLContext();
+      SSLContext sslContext = sslConfig.createSSLContext();
+
+      // if logging debug is enabled, we print the truststore entries
+      if (logger.isDebugEnabled()) {
+        this.logTrustStore();
+      }
+
+      return sslContext;
     } catch (IllegalStateException e) {
       throw new IllegalStateException(e.getCause().getMessage(), e);
     }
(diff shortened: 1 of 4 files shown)


symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -240,6 +253,21 @@ protected SSLContext createSSLContext() {
       sslConfig
           .trustStoreBytes(trustStoreBytes)
           .trustStorePassword(trustStorePassword);
+      // if logging debug is enabled, we print the truststore entries
+      if (logger.isDebugEnabled()) {
+        final KeyStore truststore;
```
symphony-soufiane, line 258: I'm wondering if this log should be added in Jersey api client as well (or only webClient one). Anyway, here I couldn't retrieve aliases as no KeyStore (trustStore) is created in this method (we use sslConfig builder too create the trustStore. So, I had to create one only to be able to retrieve the aliases and I'm not sure this is the best way to do it. Any suggestion/improvement is welcome!

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -240,6 +253,21 @@ protected SSLContext createSSLContext() {
...
+        try {
+          truststore = KeyStore.getInstance(TRUSTSTORE_FORMAT);
+          truststore.load(new ByteArrayInputStream(trustStoreBytes), trustStorePassword.toCharArray());
+          final List<String> aliases = Collections.list(truststore.aliases());
+          logger.debug("Your custom truststore contains {} entries :", aliases.size());
+          for (String alias : aliases) {
+            logger.debug("# {}", alias);
+          }
+        } catch (KeyStoreException | NoSuchAlgorithmException | CertificateException | IOException e) {
+          e.printStackTrace();
```
symphony-youri, line 268: We usually never want to use printStackTrace because it uses stderr instead of a logger so we don't have control over where the logs ended up

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -240,6 +253,21 @@ protected SSLContext createSSLContext() {
       sslConfig
           .trustStoreBytes(trustStoreBytes)
           .trustStorePassword(trustStorePassword);
+      // if logging debug is enabled, we print the truststore entries
```
symphony-youri, line 256: let's make it a separate method

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -34,6 +45,8 @@
 @API(status = API.Status.STABLE)
 public class ApiClientBuilderJersey2 implements ApiClientBuilder {
 
+  private static final Logger logger = LoggerFactory.getLogger(ApiClientBuilderJersey2.class);
```
symphony-youri, line 48: don't we use @Slf4j (lombok) in the BDK?

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -34,6 +45,8 @@
...
+  private static final String TRUSTSTORE_FORMAT = "JKS";
```
symphony-youri, line 49: usually an blank line is put to separate static / non static fields

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-youri, line 258: the variable can be declared and assigned inside the try/catch block

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -240,6 +253,21 @@ protected SSLContext createSSLContext() {
...
+      // if logging debug is enabled, we print the truststore entries
+      if (logger.isDebugEnabled()) {
+        final KeyStore truststore;
+        try {
+          truststore = KeyStore.getInstance(TRUSTSTORE_FORMAT);
+          truststore.load(new ByteArrayInputStream(trustStoreBytes), trustStorePassword.toCharArray());
+          final List<String> aliases = Collections.list(truststore.aliases());
+          logger.debug("Your custom truststore contains {} entries :", aliases.size());
+          for (String alias : aliases) {
+            logger.debug("# {}", alias);
```
symphony-youri, line 265: here we don't have much context on this log line and logs are often grepped/searched line by line
so I would probably just log for each alias something like:
`Loading {alias} from truststore'

and drop the line with the size

symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientBuilderWebClient.java
```diff
@@ -207,6 +209,15 @@ protected SslContext createSSLContext() {
           keyManagerFactory.init(keyStore, this.keyStorePassword.toCharArray());
           builder.keyManager(keyManagerFactory);
         }
+
+        // if logging debug is enabled, we print the truststore entries
```
symphony-youri, line 213: and if we can factor this if we make it a method and move it to the http-api module

symphony-bdk-http/symphony-bdk-http-webclient/src/test/java/com/symphony/bdk/http/webclient/ApiClientBuilderWebClientTest.java
```diff
@@ -37,13 +46,26 @@ void setUp() throws IOException {
...
     ApiClient apiClient = builder.build();
 
     assertEquals(apiClient.getClass(), ApiClientWebClient.class);
     assertEquals(apiClient.getBasePath(), "test-base-path");
+
+    // assert logs about truststore entries
+    List<ILoggingEvent> logsList = listAppender.list;
+    assertFalse(logsList.isEmpty(), "The list of log entries should not be empty");
+    assertNotNull(logsList.get(0), "At least on entry in the list, should not be null");
+    assertEquals("Your custom truststore contains {} entries :", logsList.get(0).getMessage(),
```
symphony-youri, line 66: the 2 asserts before are redundant IMHO as the test would fail if there is no entry

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-youri: We mostly use the jersey api client if I'm not mistaken

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-soufiane: I will replace it by throwing the runtime exception, in next commit.

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-soufiane: This is what used in ApiClientBuilderWebClient, should I change both of them to use lombok slf4j?

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-soufiane: I totally agree. However, It was done in Legacy in this way (see https://github.com/finos/symphony-bdk-java/pull/118/files#diff-5469b4312b60fd6917788efa41798433380dcc498ba75f7207c806da64b4f558R136)

Do we want to change logging to what you suggest?

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -267,20 +262,29 @@ protected SSLContext createSSLContext() {
           .keyStorePassword(keyStorePassword);
     }
     try {
-      return sslConfig.createSSLContext();
+      SSLContext sslContext = sslConfig.createSSLContext();
+
+      // if logging debug is enabled, we print the truststore entries
```
symphony-youri, line 267: the code says the exact same thing as the comment :)

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
```diff
@@ -248,9 +262,32 @@ protected SSLContext createSSLContext() {
...
+      throw new IllegalStateException(e.getCause().getMessage(), e);
+    }
+  }
+
+  private void logTrustStore() {
+    try {
+      if (isNotEmpty(trustStoreBytes) && isNotEmpty(trustStorePassword)) {
+        final KeyStore truststore = KeyStore.getInstance(TRUSTSTORE_FORMAT);
+        truststore.load(new ByteArrayInputStream(trustStoreBytes), trustStorePassword.toCharArray());
+        final List<String> aliases = Collections.list(truststore.aliases());
```
symphony-youri, line 283: can't we factor this code between the 2 implementation of the http client?

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-youri: I would change it (my personal take) (given legacy and 2.0 are very different with very little compatibility)

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientBuilderJersey2.java
symphony-soufiane: Will be done in next commit

symphony-soufiane:
ok to test

symphony-youri (approved):

Issue #515, opened 2021-05-14
Unable to create SlashHandler with same value, different @mention


symphony-soufiane:
### Steps to Reproduce:
1. Create first slash command "/test" not requiring bot's mention
2. Create another slash command "/test" which is requiring bot's mention
3. Execute both commands

### Expected Result:
When the bot is mentioned, the second one should be executed. In the other case, the first one should be executed.

### Actual Result:
Only the last one registered is executed (the other one is unsubscribed)

### Additional Context:
```
@Slash(value = "/gif")
  public void onSlashGif(CommandContext context) {
    this.messageService.send(context.getStreamId(), Message.builder().template(this.template).build());
  }

  @Slash(value = "/gif", mentionBot = false)
  public void onSlashGifNoMention(CommandContext context) {
    onSlashGif(context);
  }
```

Issue #511, opened 2021-05-04
Add Accessible Command Arguments to Slash Activity


ystan-:
### Description of Problem:
Slash commands that perform more complex tasks are usually accompanied with additional arguments behind the command itself. An example would be `/buy $GOOG 2mm` to represent a buy order for the ticker GOOG with a quantity of 2mm.

Using the BDK's `@Slash` annotation, it is currently not straightforward to extract the list of arguments.
An example of what a developer would currently do is:
- Autowire the SessionService
- Create an `@Slash` annotated method with a `.*` suffix in the command pattern
- Use SessionService to get the bot's own display name
- Extract the message text, sub-string it by the display name + 2 (for @ and space)
- Sub-string it further by the index of the first space
  - Assuming the command is a single-word here, since we have no access to the annotation value from the CommandContext
- Split the result by spaces

```java
@Autowired
private final SessionService sessionService;

@Slash("/buy .*")
public void onBuy(CommandContext context) {
  String displayName = sessionService.getSession().getDisplayName();
  String messageText = context.getTextContent().trim().substring(displayName.length() + 2);
  String[] arguments = messageText.substring(messageText.indexOf(" ") + 1).split(" ");
  String ticker = arguments[0];
  String quantity = arguments[1];
  // ... also remember to handle case without @mention prefix
}
```

### Potential Solutions:
In order of preference:
1. Support a Spring MVC-like convention when declaring `@Slash`
```java
@Slash("/buy {ticker} {quantity}")
public void onBuy(
  CommandContext context,
  @CommandArgument String ticker,
  @CommandArgument String quantity
) { ... }
```
2. Add a convenience method to the context object itself: e.g. `CommandContext::getArguments()`
3. Add a convenience class/service to extract the arguments like what was done for `PresentationMLPaser::getTextContent(String)`

symphony-youri:
To help implement it, [AntPathMatcher](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html) might be useful

symphony-elias:
Discussion on a PoC here: #595

symphony-elias:
What we are thinking of actually implementing:

**Phase 1**
* Command pattern format:
`"/command {myarg} {@mention} {$cashtag} {#hashtag}"`
All arguments are Strings by default.
* Whitespace is a strong separator, two arguments must be separated by a space.
Patterns like `"/command {arg1}{arg2}"` will be rejected
[...]

PR #516, opened 2021-05-14
#515 Unable to create SlashHandler with same value, different @mention

symphony-soufiane:
### Ticket
[#515 ](https://github.com/finos/symphony-bdk-java/issues/515)

### Description
Previously, we were not able to register two slash commands having the same name. This was done in order to avoid registering duplicates. However, two slash commands having the same name are not necessarily duplicates as they can differ in bot's mention required property : In fact, we can have one command requiring bot's mention while the second one is not requiring it. In this case, both of them should be registered. 

The issue was coming from the fact that we only relied on command's name to check if it is not already present in the registry. The fix was to check the bot's mention required property as well.

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
@@ -56,7 +56,8 @@ public class ActivityRegistry {
       Optional<AbstractActivity<?, ?>> act = this.activityList.stream()
           .filter(a -> a.getInfo().type().equals(ActivityType.COMMAND)
               && a.getInfo().name() != null
-              && a.getInfo().name().equals(activity.getInfo().name()))
+              && a.getInfo().name().equals(activity.getInfo().name())
+              && a.getInfo().requiresBotMention() == activity.getInfo().requiresBotMention())
           .findFirst();
 
       act.ifPresent(abstractActivity -> {
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
@@ -56,12 +56,14 @@ public class ActivityRegistry {
   private void preProcessActivity(AbstractActivity<?, ?> activity) {
 
     Optional<AbstractActivity<?, ?>> act = this.activityList.stream()
-        .filter(a -> a.getInfo().equals(activity.getInfo()))
+        .filter(a -> a.equals(activity))
         .findFirst();
 
     act.ifPresent(abstractActivity -> {
       abstractActivity.bindToRealTimeEventsSource(this.datafeedLoop::unsubscribe);
       this.activityList.remove(abstractActivity);
+      log.debug("One activity '{}' has been removed/unsubscribed in order to be replaced",
+          abstractActivity.getInfo().name());
     });
 
     // a command activity (potentially) needs the bot display name in order to parse the message text content
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
@@ -1,6 +1,8 @@
 package com.symphony.bdk.core.activity;
 
 import com.symphony.bdk.core.activity.command.CommandActivity;
+import com.symphony.bdk.core.activity.command.SlashCommand;
+import com.symphony.bdk.core.activity.command.HelpCommand;
 import com.symphony.bdk.core.service.datafeed.DatafeedLoop;
 import com.symphony.bdk.gen.api.model.UserV2;
 
@@ -14,6 +16,10 @@ import java.util.Optional;
 /**
  * This class allows to bind an {@link AbstractActivity} to the Real Time Events source, or Datafeed.
  * It also maintains the list of registered activities.
+ * <p>
+ * If an activity to be registered is already existing in the registry, then the old one will be replaced.
+ * In case of an activity of type {@link SlashCommand}, it will replace the old one if this latter has the same name and both require bot mention (or both don't).
+ * If the activity has /help as name, then it will replace {@link HelpCommand} if it is already registered.
  */
 @Slf4j
 @API(status = API.Status.STABLE)
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
@@ -1,7 +1,6 @@
 package com.symphony.bdk.core.activity;
 
 import com.symphony.bdk.core.activity.command.CommandActivity;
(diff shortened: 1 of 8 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
```diff
@@ -27,4 +27,7 @@
   /** Description of the activity (can contain multiple lines) */
   private String description;
 
+  /** Whether the bot mention is required or not */
+  private boolean requiresBotMention;
```
symphony-youri, line 31: to me this looks like this is specific to a command activity only and should not belong here

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/ActivityRegistryTest.java
```diff
@@ -58,4 +58,37 @@ void shouldRegisterActivity() {
...
+    final CommandActivity<?> actMentionRequired = new TestCommandActivity(true);
+    final CommandActivity<?> actMentionNotRequired = new TestCommandActivity(false);
+
+    assertTrue(this.registry.getActivityList().isEmpty(), "Registry must be empty");
+
+    this.registry.register(actMentionRequired);
+    this.registry.register(actMentionNotRequired);
+
+    verify(this.datafeedService, times(2)).subscribe(any(RealTimeEventListener.class));
+    verify(this.datafeedService, times(0)).unsubscribe(any(RealTimeEventListener.class));
```
symphony-youri, line 73: never() can be used instead of times(0)

symphony-youri (commented):
I would prefer to throw an exception if someones attempt to register an activity twice, but this won't be possible I think because we let it happen to override default activities such as the help command

thinking of it the existing com.symphony.bdk.core.activity.ActivityRegistry#preProcessActivity is casting to CommandActivity which is probably wrong

I would then prefer to reconsider that as part of https://github.com/finos/symphony-bdk-java/issues/511 to be generic and decide wether we consider the arguments of a slash command to be part of the command's identity (or just use the /... part).


also the description of the PR should be part of the commit message to make it easier to retrieve when reading git's history

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -56,7 +56,8 @@ private void preProcessActivity(AbstractActivity<?, ?> activity) {
       Optional<AbstractActivity<?, ?>> act = this.activityList.stream()
```
symphony-elias, line 56: What we could do is: remove the `instanceof` check, implement an `equals` method in the `ActivityInfo` and have a unique key field in the Activity Info (which will uniquely identify the activity).

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
symphony-elias: As discussed, we could have an arbitrary key uniquely identifying the activity, which could be a string, a list or an object whose actual type would depend on the activity.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
```diff
@@ -28,6 +30,19 @@
   private String description;
 
   /** Whether the bot mention is required or not */
-  private boolean requiresBotMention;
+  private Object uniqueObject;
```
symphony-youri, line 33: We probably want to avoid generic object so we can rely on static typing to prevent errors

Would defining equals/hashcode for activities be a way to achieve what we want?
each activity would have access to its field (whereas activityinfo is just the info and does not hold anything specific to a particular activity)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -44,31 +49,29 @@ public void register(final AbstractActivity<?, ?> activity) {
...
 
   private void preProcessActivity(AbstractActivity<?, ?> activity) {
 
+    Optional<AbstractActivity<?, ?>> act = this.activityList.stream()
+        .filter(a -> a.getInfo().equals(activity.getInfo()))
+        .findFirst();
+
+    act.ifPresent(abstractActivity -> {
+      abstractActivity.bindToRealTimeEventsSource(this.datafeedLoop::unsubscribe);
+      this.activityList.remove(abstractActivity);
```
symphony-youri, line 64: let's add a debug log here to indicate we are replacing an existing activity this could help bot's developer to understand what is going on

symphony-soufiane: Will be added in next commit

symphony-soufiane:
ok to test

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/HelpCommand.java
```diff
@@ -68,4 +68,17 @@ protected ActivityInfo info() {
         .name(HELP_COMMAND)
         .description(DEFAULT_DESCRIPTION);
   }
+
+  @Override
+  public boolean equals(Object o) {
```
symphony-youri, line 73: if equals is defined, hashcode must be too (https://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/model/ActivityInfo.java
```diff
@@ -27,4 +29,18 @@
...
   private String description;
 
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) { return true; }
+    if (o == null || getClass() != o.getClass()) { return false; }
+    ActivityInfo that = (ActivityInfo) o;
+
+    return this.hashCode() == that.hashCode();
```
symphony-youri, line 39: mm that looks odd, I don't remember seeing equality being implemented based on hashcode given there is a chance for collision I guess
Intellij can generate it for you

symphony-soufiane: We don't need it anymore in AcitivityInfo, It will be removed in next commit

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/SlashCommand.java
```diff
@@ -105,11 +107,29 @@ protected ActivityInfo info() {
...
+        .description(this.buildCommandDescription());
   }
 
   @Override
   protected CommandContext createContextInstance(V4Initiator initiator, V4MessageSent event) {
     return new CommandContext(initiator, event);
   }
+
+  private String buildCommandDescription() {
+    return this.requiresBotMention ? this.description + " (mention required)"
```
symphony-youri, line 119: I have a feeling we might come back to it in the future with users that don't want this part to appear in the description :D

symphony-youri (approved):

Issue #517, opened 2021-05-19
List users method does not handle bad path


symphony-soufiane:
## Bug Report
The [list users method ](https://github.com/finos/symphony-bdk-java/blob/4e32de3b38e01edb1844dd3658645f1ab7877936/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java#L130)does not handle bad paths well. it currently throws a null pointer. 

and i would expect either:
a deliberate exception (the legacy sdk throws NoContentException)
or an empty collection (since the input is a list, the returned list could have less entries if any of them are non-existent)

### Steps to Reproduce:
`execute bdk.users().listUsersByEmails(List.of("x@x.com"), true, true);`
### Expected Result:
Throw "NoContentException" (like legacy sdk) or return an empty collection (since the input is a list, the returned list could have less entries if any of them are non-existent).

### Actual Result:
Throws null pointer exception.

PR #518, opened 2021-05-19
#517 List users method does not handle bad path.

symphony-soufiane:
### Ticket
[#517](https://github.com/finos/symphony-bdk-java/issues/517)

### Description
User service methods to search users cause null pointer when no user is found.
The endpoint /v3/users to search users return null with 204 (no content) http code when no user is found.
This endpoint is used in user service to search users by ids, emails and usernames.

The null pointer was comming from User service methods that call getUsers() on the returned object, which causes NP when no user is found.
There were two options to fix this :
- throw an exception when no user is found
- return an empty list

The second option was chosen to fix this bug. Thus, User service returns an empty list when no user found.

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
@@ -110,12 +110,7 @@ public class UserService implements OboUserService, OboService<OboUserService> {
     String uids = uidList.stream().map(String::valueOf).collect(Collectors.joining(","));
     V2UserList v2UserList = executeAndRetry("searchUserByIds",
         () -> usersApi.v3UsersGet(authSession.getSessionToken(), uids, null, null, local, active));
-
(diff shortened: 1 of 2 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -165,6 +191,11 @@ public OboUserService obo(AuthSession oboSession) {
     String usernames = String.join(",", usernameList);
     V2UserList v2UserList = executeAndRetry("searchUserByUsernames",
         () -> usersApi.v3UsersGet(authSession.getSessionToken(), null, null, usernames, true, null));
+
+    if (v2UserList == null || v2UserList.getUsers() == null) {
```
symphony-youri, line 195: as v2UserList is of the same type we could probably factor this logic in a `getUsersOrEmpty(v2UserList)` method

symphony-youri (approved):

symphony-mariacristina (approved):
LGTM!

Issue #521, opened 2021-05-25
Upgrade jersey dependencies to 2.34 or newer


symphony-soufiane:
## Feature Request
Upgrade jersey dependencies

### Description of Problem:
Both legacy and BDK are still using 2.32
Recommended fix is to upgrade to 2.34 or newer.

PR #522, opened 2021-05-25
PLAT-10897 Upgraded jersey dependency to 2.34

symphony-soufiane:
### Ticket
[#521](https://github.com/finos/symphony-bdk-java/issues/521)

### Description
Upgraded jersey dependency to 2.34


symphony-youri (approved):
Maybe we could give the jersey bom a shot (https://stackoverflow.com/questions/22870422/what-is-the-purpose-of-including-the-jersey-bom-import-scoped-dependency-in-a-je) if it avoids repeating the versions

Issue #523, opened 2021-05-25
Upgrade Jersey and Commons dependancies in Legacy


symphony-soufiane:
## Feature Request
Legacy BDK is still using Jersey 2.32 and Commons 2.6

symphony-youri:
Fixed by https://github.com/finos/symphony-bdk-java/pull/524

PR #524, opened 2021-05-25
PLAT-10897 PLAT-10898 fix Jersey and Commons dependencies in Legacy

symphony-soufiane:
### Ticket
[#523](https://github.com/finos/symphony-bdk-java/issues/523)

### Description
Upgraded jersey dependency to 2.34 and Commons to 2.7

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,7 +88,7 @@ subprojects {
 
             implementation 'commons-beanutils:commons-beanutils:1.9.4'
             implementation 'commons-codec:commons-codec:1.15'
-            implementation 'commons-io:commons-io:2.7'
+            implementation 'commons-io:commons-io:2.8.0'
 
             implementation 'io.github.resilience4j:resilience4j-bulkhead:1.4.0'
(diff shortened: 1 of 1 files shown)


symphony-youri:
Do we want to keep updating the legacy BDK? Were there specific vulnerabilities associated to those dependencies?

symphony-soufiane:
> Do we want to keep updating the legacy BDK? Were there specific vulnerabilities associated to those dependencies?

Some of our customers are still using the legacy BDK, that's why we were asked to do the fix here.

symphony-youri (approved):

anthony-symphony:
@symphony-soufiane any particular reason we are only going to commons-io 2.7 instead of going straight to 2.8?

symphony-soufiane:
> @symphony-soufiane any particular reason we are only going to commons-io 2.7 instead of going straight to 2.8?

@anthony-symphony it was corrected to 2.8.0 in a later commit.

PR #541, opened 2021-06-30
Update OpenAPI generator

symphony-youri:
To make it work with recent versions of Gradle.
Templates had to be updated using the latest changes from OpenAPI
generator repository.

Also better define task dependencies to avoid Gradle warnings.

---

Fix slow load balancing test

By default retries are performed, make it retry only once.
Also clean up Sonar/IDEA warnings.


symphony-soufiane (approved):

symphony-mariacristina (approved):
LGTM!

symphony-elias (approved):

Issue #546, opened 2021-07-09
Load both custom and system trustore in the Java BDKs


symphony-mariacristina:
### Steps to Reproduce:
If we use a custom trustore in the configuration of the bot, it currently replace the system default one. This cause some failure when trying to start the bot in an environment which contains system certificates that should be recognised by default.

### Expected Result:
Bot should never throw an SSL certificate error when trying to use default system certificates

**Fix to be included in both BDK 2.0 and legacy one**

symphony-mariacristina:
PR for legacy https://github.com/finos/symphony-bdk-java/pull/547

PR #547, opened 2021-07-09
Load system truststore in the SSL Context

symphony-mariacristina:
### Ticket
#546 
### Description
Goal of this pr is to load both custom and system truststore while creating the SSL context.

### Dependencies

### Checklist


symphony-youri (approved):
In term of unit test it could be done with mockserver I guess (like with the same client try to hit https://google.com and the local server using a custom cert)

PR #554, opened 2021-07-30
Keep DF loop running for UnknownHostException

symphony-youri:
The datafeed loop could stop if DNS related exception were raised (which
could happen because of network glitches or after an outage).

Similar to the timeoutout or connect error we prefer to keep the DF loop
running and to retry.

Note that for DFv2 on startup a DNS error will prevent the DF loop to
start (same as a connect error).
For DFv1 a DNS error will still allow the DF loop to start (if the DF id
already exists) this is consistent with the existing behavior for
connect errors.

Fixes PLAT-11178


symphony-elias (approved):
LGTM

symphony-elias:
Does it work for webclient as well ?

symphony-mariacristina (approved):

symphony-youri:
> Does it work for webclient as well ?

With webclient SearchDomainUnknownHostException is raised and it is a subclass of UnknownHostException so we should be good

Issue #561, opened 2021-08-06
Make ApplicationEventMulticaster bean configurable


yinan-symphony:
## Feature Request
Make the ApplicationEventMulticaster bean in BdkDatafeedConfig.java configurable, so the bdk client application could choose to use or not use the with the executor to run the event consumer asynchronously.

### Description of Problem:
Running the application event consumer asynchronously is not always preferable, this feature will give the a better control on that to BDK client.

### Potential Solutions:
Add a new bdk.datafeed.async boolean property, when it is true, this bean is created, otherwise not. True is the default value to keep the backward compatibility.

PR #562, opened 2021-08-06
Feature-561: Make ApplicationEventMulticaster bean configurable

yinan-symphony:
Running the application event consumer asynchronously is not always preferable,
this feature will give the a better control on that to BDK client.

Example:
bdk:
    datafeed:
        async: false

### Ticket
[561](https://github.com/finos/symphony-bdk-java/issues/561)

### Description
Please put here the intent of your pull request.

### Dependencies

### Checklist
- [ ] Referenced a ticket in the PR title and in the corresponding section

### DIFF
--- a/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
+++ b/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
@@ -70,6 +70,7 @@ public class BdkDatafeedConfig {
    * Allows to publish application {@link RealTimeEvent} asynchronously from {@link RealTimeEventsDispatcher}.
    */
   @Bean(name = "applicationEventMulticaster")
+  @ConditionalOnProperty(value = "bdk.datafeed.event.async", havingValue = "true", matchIfMissing = true)
   public ApplicationEventMulticaster simpleApplicationEventMulticaster() {
     final SimpleApplicationEventMulticaster eventMulticaster = new SimpleApplicationEventMulticaster();
     eventMulticaster.setTaskExecutor(new SimpleAsyncTaskExecutor());
--- a/docs/spring-boot/core-starter.md
+++ b/docs/spring-boot/core-starter.md
@@ -223,6 +223,15 @@ public class RealTimeEvents {
 }
 ```
 
+By default, the RealTimeEvents are going to be processed asynchronously in the listeners, in case this is not the preferred behavior, one
+can deactivate it by updating the application.yaml file as
+```yaml
+bdk:
+    datafeed:
+        event:
+            async: false
+```
+
 ## Inject Services
 The Core Starter injects services within the Spring application context:
 ```java


yinan-symphony:
Unfortunately, i did not find a way to test the solution, as the TaskExecutor getter method is not accessible.

imadamahmid1 (approved):

symphony-elias:
@yinan-symphony we agree on the principle. But please add a corresponding test in SymphonyBdkAutoConfigurationTest

yinan-symphony:
@symphony-elias as you might have seen my comment above, i was trying to add a unit test using spring context runner, but the taskExecutor of the ApplicationEventMulticaster is not accessible, i cannot test it, and by default ApplicationEventMulticaster is a bean from spring context, in SymphonyBdkAutoConfiguration, we try to custom this bean by adding the taskExecutor, so i do not have an idea how to create a unit test to make it sense. @symphony-thibault , any idea? @symphony-youri  ?

symphony-youri:
We could get the bean in the test and then use it with a custom listener (calling multicastEvent()) that checks it is executed in the same thread or not
Maybe a bit cumbersome, I would be ok without the test

We should update the documentation here https://github.com/finos/symphony-bdk-java/blob/a0c2b8c674bf1b4a27e6b284fdbb052b0467c095/docs/spring-boot/core-starter.md#L162 too to introduce this new parameter and maybe give use cases for its usage. What are they by the way? Do we have bots that are running with a single instance and processing events one by one?

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
```diff
@@ -70,6 +70,7 @@ public DatafeedAsyncLauncherService datafeedAsyncLauncherService(final DatafeedL
    * Allows to publish application {@link RealTimeEvent} asynchronously from {@link RealTimeEventsDispatcher}.
    */
   @Bean(name = "applicationEventMulticaster")
+  @ConditionalOnProperty(value = "bdk.datafeed.event.async", havingValue = "true", matchIfMissing = true)
```
symphony-thibault, line 73: Don't we need to reflect this property in https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/SymphonyBdkCoreProperties.java to make it properly typed ?

yinan-symphony: @symphony-thibault , today's "bdk.datafeed.enabled" property is not reflected to this SymphonyBdkCoreProperties.java either, i was thinking this was on purpose as it is a property used in one place in the autoconfig class only. Let me know if you want to do it.

yinan-symphony:
> We could get the bean in the test and then use it with a custom listener (calling multicastEvent()) that checks it is executed in the same thread or not
> Maybe a bit cumbersome, I would be ok without the test
> 
> We should update the documentation here
> 
> https://github.com/finos/symphony-bdk-java/blob/a0c2b8c674bf1b4a27e6b284fdbb052b0467c095/docs/spring-boot/core-starter.md#L162
> 
> too to introduce this new parameter and maybe give use cases for its usage. What are they by the way? Do we have bots that are running with a single instance and processing events one by one?

i was looking for a document on this BdkDatafeedConfig, there is basically nothing yet, i can definitely add this new config. This ApplicationEventMulticaster is applied globally per spring container, the asyn processing makes a unexpected result in case some data base modules use event listeners to add extra info to the DB data models before persisting them into DB, typically use case is the SSBS mongo module.

yinan-symphony:
@symphony-youri doc updated

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/config/BdkDatafeedConfig.java
symphony-thibault: Ok ok @yinan-symphony 👍  Let's keep it as it is for now, especially since the property is now documented

symphony-youri:
> > We could get the bean in the test and then use it with a custom listener (calling multicastEvent()) that checks it is executed in the same thread or not
> > Maybe a bit cumbersome, I would be ok without the test
> > We should update the documentation here
> > https://github.com/finos/symphony-bdk-java/blob/a0c2b8c674bf1b4a27e6b284fdbb052b0467c095/docs/spring-boot/core-starter.md#L162
> > 
> > too to introduce this new parameter and maybe give use cases for its usage. What are they by the way? Do we have bots that are running with a single instance and processing events one by one?
> 
> i was looking for a document on this BdkDatafeedConfig, there is basically nothing yet, i can definitely add this new config. This ApplicationEventMulticaster is applied globally per spring container, the asyn processing makes a unexpected result in case some data base modules use event listeners to add extra info to the DB data models before persisting them into DB, typically use case is the SSBS mongo module.

What's the unexpected result?

symphony-youri (approved):
Ok for me, I would like to better understand the use case for it though (hence the last question about the unexpected result)

symphony-thibault (approved):

PR #489, opened 2021-04-14
#482 traceId is set for each new DF events, Jersey2 client logs request execution time

symphony-thibault:
### Ticket
Closes #482

### Description
When DF loop dispatches events, a new `DistributedTracingContext`'s traceId is set using the event ID. Here's an example of logs when DF loop receives and dispatches an event: 
```
09:18:23.380 [main] oxRa0S DEBUG com.symphony.bdk.requests.outgoing - status=200, url=https://devx3.symphony.com:443/agent/v5/datafeeds/9a25e271dff12046052748f1d58aa2b3_f_2915e/read, time=8194
09:18:23.571 [main] 4v4RuD DEBUG c.s.b.c.s.d.i.AbstractDatafeedLoop - Before dispatching 'MESSAGESENT' event to listener 78816013 (hashcode)
09:18:24.579 [main] 4v4RuD DEBUG com.symphony.bdk.requests.outgoing - status=200, url=https://devx3.symphony.com:443/agent/v4/stream/cJmrxcLsZIxbsONbbAnjh3___oc5WHjzdA/message/create, time=981
09:18:24.585 [main] 4v4RuD DEBUG c.s.b.c.s.d.i.AbstractDatafeedLoop - 'MESSAGESENT' event successfully dispatched to listener 78816013 (hashcode)
09:18:24.585 [main] oxRa0S DEBUG c.s.b.c.s.d.impl.DatafeedLoopV2 - Reading datafeed events from datafeed 9a25e271dff12046052748f1d58aa2b3_f_2915e
09:18:24.777 [main] oxRa0S DEBUG com.symphony.bdk.requests.outgoing - status=200, url=https://devx3.symphony.com:443/agent/v5/datafeeds/9a25e271dff12046052748f1d58aa2b3_f_2915e/read, time=191
09:18:24.778 [main] oxRa0S DEBUG c.s.b.c.s.d.impl.DatafeedLoopV2 - Reading datafeed events from datafeed 9a25e271dff12046052748f1d58aa2b3_f_2915e
09:18:24.983 [main] oxRa0S DEBUG com.symphony.bdk.requests.outgoing - status=200, url=https://devx3.symphony.com:443/agent/v5/datafeeds/9a25e271dff12046052748f1d58aa2b3_f_2915e/read, time=204
[...]

PR #569, opened 2021-08-27
ApiClientJersey2 initialize random traceId if not already set

symphony-thibault:
PR https://github.com/finos/symphony-bdk-java/pull/489 has introduced a regression where the `ApiClientJersey2` stopped initializing the `traceId` value if not already set: 
Screen Shot 2021-08-27 at 17 12 03

Here we revert this change but we now ensure that the `traceId` is cleared in order to not pollute subsequent logs.

### DIFF
--- a/symphony-bdk-http/symphony-bdk-http-jersey2/build.gradle
+++ b/symphony-bdk-http/symphony-bdk-http-jersey2/build.gradle
@@ -37,5 +37,7 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'ch.qos.logback:logback-classic'
     testImplementation 'org.mock-server:mockserver-netty'
+    testImplementation 'org.mockito:mockito-core'
+    testImplementation 'org.mockito:mockito-junit-jupiter'
 }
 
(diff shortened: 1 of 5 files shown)


symphony-youri:
By clearing the trace id here don't we loose it in the logs produced for each API call (with the status/response time)?

It could be implemented as a Closeable resource too (but it comes at the cost of creating a new object instead of passing this boolean around).

Should we have this same behavior (with the clearing of MDC) in ApiClientWebClient too?

In the end isn't it wrong that we set the trace id in com.symphony.bdk.core.service.datafeed.impl.DatafeedLoopV2#start ?
It is going to stay the same for all the time the DF loop is running

symphony-youri (approved):
The behavior with the clearTraceId can probably be tested with unit tests

symphony-thibault:
> The behavior with the clearTraceId can probably be tested with unit tests

@symphony-youri I applied your suggestion (cf. https://github.com/finos/symphony-bdk-java/pull/569/commits/a1e7c0c4a45e42ff73776e9096732b3a2708c224)

symphony-youri (approved):

PR #572, opened 2021-09-09
Fixed characters escape in PresentationMLParser

symphony-soufiane:
PresentationMLParser getTextContent method was unescaping content characters before parsing it which fails when the message contains messageML tags for instance. In this commit, we changed operations order to parse the message first with the escaped characters, then unescape it in a second time

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/PresentationMLParser.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/PresentationMLParser.java
@@ -32,13 +32,13 @@ public class PresentationMLParser {
    * @param trim            flag if we want to trim the text result
    * @return the message text content extracted from the given PresentationML
    */
-  public static String getTextContent(String presentationML, Boolean trim) throws PresentationMLParserException {
+  public static String getTextContent(String presentationML, boolean trim) throws PresentationMLParserException {
     try {
-      String escapedPresentationML = StringEscapeUtils.unescapeHtml4(presentationML);
       final Document doc = LOCAL_BUILDER.get().parse(
-          new ByteArrayInputStream(escapedPresentationML.getBytes(StandardCharsets.UTF_8)));
+          new ByteArrayInputStream(presentationML.getBytes(StandardCharsets.UTF_8)));
       String textContent = doc.getChildNodes().item(0).getTextContent();
-      return trim ? textContent.trim() : textContent;
+      String escapedPresentationML = StringEscapeUtils.unescapeHtml4(textContent);
+      return trim ? escapedPresentationML.trim() : escapedPresentationML;
     } catch (SAXException | IOException e) {
       throw new PresentationMLParserException(presentationML, "Failed to parse the PresentationML", e);
     }
--- a/symphony-bdk-core/src/test/java/com/symphony/bdk/core/util/PresentationMLParserTest.java
+++ b/symphony-bdk-core/src/test/java/com/symphony/bdk/core/util/PresentationMLParserTest.java
@@ -3,23 +3,36 @@ package com.symphony.bdk.core.util;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 import com.symphony.bdk.core.service.message.exception.PresentationMLParserException;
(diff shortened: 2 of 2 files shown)


symphony-bdk-core/src/test/java/com/symphony/bdk/core/util/PresentationMLParserTest.java
```diff
@@ -3,23 +3,36 @@
...
+class PresentationMLParserTest {
+
+  static Stream<Arguments> validPresentationMLs() {
+    return Stream.of(
+        arguments(
+            "<div data-format=\"PresentationML\" data-version=\"2.0\">\n<a href=\"http://www.symphony.com\">This is a link to Symphony's Website</a>\n</div>",
+            "This is a link to Symphony's Website"),
+        arguments("<div data-format=\"PresentationML\" data-version=\"2.0\"> <p>/test &lt;/messageML&gt;</p> </div>",
+            "/test </messageML>"),
+        arguments("<div data-format=\"PresentationML\" data-version=\"2.0\">Hello&#xA0;World</div>", "Hello World"));
```
symphony-soufiane, line 27: Please note that here I changed &NBSP; with its hexadecimal notation &#xA0; 
The test was failing because &NBSP; entity is not predefined in the doctype. (see https://stackoverflow.com/questions/13012327/error-parsing-page-xhtml-error-tracedline-42-the-entity-nbsp-was-referenc) 

Is it a breaking change or not?

symphony-soufiane:
ok to test

symphony-bdk-core/src/test/java/com/symphony/bdk/core/util/PresentationMLParserTest.java
symphony-elias: I wouldn't say so. To be sure, please run the same unit tests without changes

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/PresentationMLParser.java
```diff
@@ -32,13 +32,13 @@
    * @param trim            flag if we want to trim the text result
    * @return the message text content extracted from the given PresentationML
    */
-  public static String getTextContent(String presentationML, Boolean trim) throws PresentationMLParserException {
+  public static String getTextContent(String presentationML, boolean trim) throws PresentationMLParserException {
```
symphony-elias, line 35: Are we sure we never put `null` as `trim` parameter ?

symphony-soufiane: Putting null there should not compile IMHO but let me put it back to Boolean. Basically I changed it to fix a sonar warning.

symphony-bdk-core/src/test/java/com/symphony/bdk/core/util/PresentationMLParserTest.java
symphony-soufiane: To make it clear: 
Before my changes, when we put the presentationML with **_Hello&nbsp;World_** the returned content was **_HelloNBSPWorld_** . Now with my changes, the presentationML should have **_Hello&#xA0;World_** instead of **_HelloNBSPWorld_**, otherwise an error is thrown. 
The error and more explanations are in stackoverflow link in my first comment above.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/PresentationMLParser.java
symphony-youri: This is breaking the API contract :)

symphony-bdk-core/src/test/java/com/symphony/bdk/core/util/PresentationMLParserTest.java
symphony-youri: I think we added this test because the client could generate such content with &nbsp;
so we are not testing the same thing now are we?

symphony-soufiane:
ok to test

symphony-soufiane:
ok to test

symphony-youri:
there is not ok to test here :)

symphony-youri:
This change probably deserves a bit of explanation
Are we sure that nbsp is the only character we have to handle?

ystan-:
there's another list of these entities in this class

https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageUtil.java#L47-L70

PR #573, opened 2021-09-10
Prevent invalid signed JWT expiration upon login

symphony-youri:
If the backend and the bot/BDK do not have their clocks synced, the
signed JWT generated to authenticate could be considered as invalid
because the expiration is considered to be more then 5 minutes by the
backend.

We use 4 minutes instead to accommodate that.

Given we create a new token for each authentication attempt this should
be fine to use less than 5 minutes (we expect the signed JWT to be used
quickly).

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/jwt/JwtHelper.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/jwt/JwtHelper.java
@@ -42,7 +42,7 @@ import java.util.Date;
 @API(status = API.Status.INTERNAL)
 public class JwtHelper {
 
-  // Expiration of the jwt, 5 minutes maximum, use a little less than that in case of clock differents
+  // Expiration of the jwt, 5 minutes maximum, use a little less than that in case of different clock skews
   public static final Long JWT_EXPIRATION_MILLIS = 240_000L;
 
 	// PKCS#8 format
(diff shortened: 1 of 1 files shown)


symphony-youri:
In the Python BDK it is already using a smaller value than 5 minutes

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/jwt/JwtHelper.java
```diff
@@ -42,8 +42,8 @@
 @API(status = API.Status.INTERNAL)
 public class JwtHelper {
 
-  // Expiration of the jwt
-  public static final Long JWT_EXPIRATION_MILLIS = 300_000L;
+  // Expiration of the jwt, 5 minutes maximum, use a little less than that in case of clock differents
```
symphony-elias, line 45: I would say "in case of different clocks" or "in case of different clock skews" ?

symphony-elias (approved):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/jwt/JwtHelper.java
symphony-youri: Indeed I meant clock differences, I fixed it 👍

symphony-thibault (approved):

PR #579, opened 2021-09-27
[beta] Support of message update endpoint

symphony-thibault:
Added new message update endpoint to `MessageService` class.
- Mentioned in markdown doc 
- Simple demo created (`MessageUpdateExampleMain.java`). I'd like make it a bit more fun so I'll modify it before merging this PR
----
message-update-show

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
@@ -290,6 +290,25 @@ public class MessageService implements OboMessageService, OboService<OboMessageS
   }
 
   /**
+   * Update an existing message. The existing message must be a valid social message, that has not been deleted.
+   *
+   * @param streamId the ID of the stream to send the message to
+   * @param messageId the ID of the message to be
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @see <a href="https://developers.symphony.com/restapi/reference#update-message-v4">Create Message v4</a>
+   */
+  public V4Message update(@Nonnull String streamId, @Nonnull String messageId, @Nonnull Message message) {
+    return this.executeAndRetry("update", messagesApi.getApiClient().getBasePath(), () ->  {
+      final String path = String.format(
+          "/v4/stream/%s/message/%s/update",
+          this.messagesApi.getApiClient().escapeString(toUrlSafeIdIfNeeded(streamId)),
+          this.messagesApi.getApiClient().escapeString(toUrlSafeIdIfNeeded(messageId))
+      );
+      return doSendFormData(path, getForm(message), new TypeReference<V4Message>() {});
+    });
+  }
+
+  /**
    * Sends a message to multiple existing streams.
    *
    * @param streamIds the list of stream IDs to send the message to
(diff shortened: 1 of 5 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -289,6 +289,26 @@ public V4Message send(@Nonnull String streamId, @Nonnull Message message) {
         () -> this.doSendMessage(streamId, message));
   }
 
+  /**
+   * Update an existing message. The existing message must be a valid social message, that has not been deleted.
+   *
+   * @param streamId the ID of the stream to send the message to
+   * @param messageId the ID of the message to be
+   * @return a {@link V4Message} object containing the details of the sent message
+   * @see <a href="https://developers.symphony.com/restapi/reference#update-message-v4">Create Message v4</a>
```
symphony-youri, line 298: Update Message v4? (in the link's name)

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/MessageUpdateExampleMain.java
```diff
@@ -0,0 +1,35 @@
...
+ * This demonstrates a basic usage of the message update feature.
+ */
+public class MessageUpdateExampleMain {
+
+  public static void main(String[] args) throws Exception {
+
+    final SymphonyBdk bdk = new SymphonyBdk(loadFromSymphonyDir("config.yaml"));
+
+    bdk.activities().register(slash("/go", false, context -> {
+      newSingleThreadExecutor().submit(() -> startTheShow(bdk, context.getStreamId()));
```
symphony-youri, line 23: why do we need the thread executor?

symphony-youri (approved):

symphony-mariacristina (approved):
Should we maybe also mention that we cannot update a message containing attachments?

symphony-thibault:
> @symphony-mariacristina - Should we maybe also mention that we cannot update a message containing attachments?

Good point, we have different solutions: 
1. throw an error if `Message#attachments` is not empty
2. create another `Message` that does not contain `attachments` and `previews` fields
3. just mention that in Javadoc

I personally prefer solution `3`, as we plan to support attachments in a near future, it will avoid us to modify the BDK once message update works with attachments.

symphony-mariacristina:
> I personally prefer solution `3`, as we plan to support attachments in a near future, it will avoid us to modify the BDK once message update works with attachments.

yes, I agree with point 3. The API is marked as experimental and will be changed anyway to support attachments.

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/MessageUpdateExampleMain.java
symphony-thibault: To void blocking the DF loop, and run multiple `/go` at the same time. But I just updated the example and removed it

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/MessageUpdateExampleMain.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @SneakyThrows
+  private static void startTheShow(SymphonyBdk bdk, String streamId) {
+
+    // just for this example, obviously!
+    Function<Integer, String> pickEmoji = index -> new String[] {
+        "cat", "dromedary_camel", "dolphin", "dog", "hamster", "goat", "panda_face", "koala", "frog", "penguin"
+    }[index];
+
+    V4Message previousMessage = bdk.messages().send(streamId, "Let's the show begin!");
```
symphony-thibault, line 40:
```suggestion
    V4Message previousMessage = bdk.messages().send(streamId, "Let the show begin!");
```

PR #591, opened 2021-11-18
Improving freemarker template engine configuration

symphony-mariacristina:
### Description
- Enforcing numbers format to not add , or . in between. Before while using numbers, freemarker was formatting them from "123456" to "123,456" this was causing issues especially when dealing with userIds.
- Enforcing freemarker to access beans object fields without the need to specify the get in the variable. Before only this definition was working ${event.getParameter().getAnother()}, now also this one will work: ${event.parameter.another}

### DIFF
--- a/symphony-bdk-template/symphony-bdk-template-freemarker/src/main/java/com/symphony/bdk/template/freemarker/FreeMarkerEngine.java
+++ b/symphony-bdk-template/symphony-bdk-template-freemarker/src/main/java/com/symphony/bdk/template/freemarker/FreeMarkerEngine.java
@@ -60,7 +60,7 @@ public class FreeMarkerEngine implements TemplateEngine {
     cfg.setLogTemplateExceptions(false);
     cfg.setWrapUncheckedExceptions(true);
     cfg.setFallbackOnNullLoopVariable(false);
-    cfg.setNumberFormat("0.######");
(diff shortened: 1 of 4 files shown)


symphony-bdk-template/symphony-bdk-template-freemarker/src/main/java/com/symphony/bdk/template/freemarker/FreeMarkerEngine.java
```diff
@@ -59,6 +60,10 @@ private static Configuration createConfiguration() {
     cfg.setLogTemplateExceptions(false);
     cfg.setWrapUncheckedExceptions(true);
     cfg.setFallbackOnNullLoopVariable(false);
+    cfg.setNumberFormat("0.######");
```
symphony-youri, line 63: Does it mean we drop the fraction parts over 6 digits?

symphony-mariacristina: Unfortunately yes. But I just found another parameter that we could use "computer". it has the same impact as this https://freemarker.apache.org/docs/ref_builtins_number.html#ref_builtin_c . Let me replace it.

symphony-youri (approved):

Issue #592, opened 2021-11-18
Enums are not flexible on new fields


symphony-mariacristina:
### Steps to Reproduce:
- Make sure pod contains SDL users
- Call Users lookup endpoint (GET /pod/v3/users/), using BDK with version  <= 2.2.1

### Expected Result:
Even if the generated model do not contain the value "SDL" in the enum as a valid account type, the BDK should still be able to parse the results.

### Actual Result:
BDK fails with the following exception:
```
java.lang.IllegalArgumentException: Unexpected value 'SDL'
at com.symphony.bdk.gen.api.model.UserV2$AccountTypeEnum.fromValue(UserV2.java:117)
```

symphony-thibault:
@symphony-mariacristina @symphony-youri the only solution I see on library side is to modify the [modelInnerEnum.mustache](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/modelInnerEnum.mustache) in order to[ enforce returning `null` instead of throwing `IllegalArgumentException` ](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/Java/modelInnerEnum.mustache#L58) if the generated code is not up-to-date.

This is clearly not the best solution, but at least it will give avoid breaking any API call in case we add new value in the future.

symphony-youri:
Given we cannot change the existing APIs/models
And that some of them are already using enums
I don't see a better solution, at least this way we might not break some of users in the future (right now all the users of the BDK would have to update anyway). They would get null enums in case new values are added but this might be ok if they don't use it.

Maybe we could add a default value to all existing enums (UNKNOWN) that could be returned instead of null but in this case I'm not sure we are not breaking compatibility anyway (https://wiki.eclipse.org/Evolving_Java-based_APIs_2 seems to say it would be ok).

Could we maybe evaluate the affected enums in the recent releases? (or is it only AccountType SDL?)

There is also https://github.com/OAI/OpenAPI-Specification/issues/1552 on this topic.

PR #593, opened 2021-11-29
#592 Generated enums don't throw exception when parsing unknown values

symphony-thibault:
Closes https://github.com/finos/symphony-bdk-java/issues/592

We want to avoid response parsing issues when API returns an unknown enum value. The `modelInnerEnum.mustache` template has been modified in order to: 
- return null
- print warning log

### DIFF
--- a/symphony-bdk-core/src/test/java/com/symphony/bdk/core/gen/GeneratedEnumTest.java
+++ b/symphony-bdk-core/src/test/java/com/symphony/bdk/core/gen/GeneratedEnumTest.java
@@ -1,0 +1,39 @@
+package com.symphony.bdk.core.gen;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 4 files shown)


symphony-elias (approved):

symphony-elias:
Could we add a test when a given endpoint returns an unknown enum value ?

symphony-youri (approved):
We might want to explain a bit more the context in the commit msg, especially the recent SDL enum change

symphony-thibault:
> Could we add a test when a given endpoint returns an unknown enum value ?

Done @symphony-elias: bf90006

symphony-thibault:
> We might want to explain a bit more the context in the commit msg, especially the recent SDL enum change

Done @symphony-youri : https://github.com/finos/symphony-bdk-java/commit/7c68ccc6ccf73e17e0cf061734ad5cc7c13f2c33

PR #595, opened 2021-12-02
Slash commands Spring like PoC

symphony-youri:
I'm just opening the PR so we can comment on it


symphony-bdk-examples/bdk-spring-boot-example/src/main/java/com/symphony/bdk/examples/spring/GifFormActivity.java
```diff
@@ -37,6 +37,11 @@ public void displayGifForm(CommandContext context) {
     this.messageService.send(context.getStreamId(), message);
   }
 
+  @Slash("/echo {one} {two}")
+  public void echo(CommandContext context, String one, String two) {
```
symphony-youri, line 41: do we start introducing @SlashParam annotations too (similar to PathParam)? To customize the mappings or have optional parameters

symphony-youri: do we support multiple types here?

symphony-elias: No, not yet. But we could. See discussion in the confluence page.
I would say specifying the fact a param is optional should be put in the command pattern and should put after all mandatory params.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivity.java
```diff
@@ -53,10 +59,23 @@ public void onMessageSent(V4Initiator initiator, V4MessageSent event) {
...
   protected void beforeMatcher(C context) {
     try {
-      context.setTextContent(
-          PresentationMLParser.getTextContent(context.getSourceEvent().getMessage().getMessage()));
+      final String textContent =
+          PresentationMLParser.getTextContent(context.getSourceEvent().getMessage().getMessage());
+      context.setTextContent(textContent);
+
+      // Fails if message contains a word beginning with @ without being an actual mention
+      List<String> mentionsText = Arrays.stream(textContent.split("\\s+")).filter(s -> s.startsWith("@")).collect(Collectors.toList());
```
yinan-symphony, line 70: what if the mention has a space in between @Yinan LIU,  it is a mention, if i understand well, this will only put @Yinan in the list of string, no?

yinan-symphony: we have the same feature already implemented in Harmony framework, waiting for this done in BDK, so that we can remove the same from HFK. a bit shame what we cannot communicate the features shared/done in one or the other.

symphony-thibault: > what if the mention has a space in between @Yinan LIU, it is a mention, if i understand well, this will only put @Yinan in the list of string, no?

Good point @yinan-symphony 👍 

@symphony-elias you might need to use [MessageParser#getMentions](https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/util/MessageParser.java#L36)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/AntPathMatcher.java
```diff
@@ -0,0 +1,965 @@
+package com.symphony.bdk.core.util;
+
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
```
symphony-thibault, line 10: Are we OK with the license?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/StringUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+
+import org.apiguardian.api.API;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.StringTokenizer;
+
+@API(status = API.Status.EXPERIMENTAL)
```
symphony-thibault, line 11: I'd even mark `INTERNAL`

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/command/SlashCommandTest.java
```diff
@@ -44,6 +47,55 @@ void testSlashCommandWithBotMentionSuccess() {
     assertTrue(handlerCalled.get());
   }
 
+  @Test
+  void testSlashCommandsWithArgs() {
```
symphony-thibault, line 51: Could we consider using [JUnit 5 Parametrized tests](https://www.baeldung.com/parameterized-tests-junit-5) here to test more use cases? (just a proposal)

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/activity/SlashCommandsWithArgs.java
```diff
@@ -0,0 +1,29 @@
...
+public class SlashCommandsWithArgs {
+
+  public static void main(String[] args) throws Exception {
+    final SymphonyBdk bdk = new SymphonyBdk(loadFromSymphonyDir("config.yaml"));
+
+    bdk.activities().register(new SlashArgumentActivity("/echo {one} {two}", true, (context, slashArgs) ->
+        bdk.messages().send(context.getStreamId(), "Args: " + slashArgs)
+    ));
+
+    bdk.activities().register(new SlashMentionArgumentActivity("/mentions {one} {two}", true, (context, slashArgs) ->
```
symphony-thibault, line 21: I'm really not in favour of having a different activity for mentions. I'd expect something like: 
```java
bdk.activities().register(new SlashCommandActivity("/echo {one} @{two} #{three} ${four}", true, (context, slashArgs) ->
        bdk.messages().send(context.getStreamId(), "Args: " + slashArgs)
)); 
```

symphony-bdk-examples/bdk-spring-boot-example/src/main/resources/application.yaml
```diff
@@ -1,10 +1,15 @@
 
 bdk:
-  host: acme.symphony.com
+  host: devx1.symphony.com
```
symphony-thibault, line 3: Needs to be reverted, locally I prefer specifying those config params from IntelliJ (e.g. `--bdk.host=devx1.symphony.com`)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivity.java
symphony-elias: Indeed, and pitfalls of this implementation have already been described here: https://perzoinc.atlassian.net/wiki/spaces/DevX/pages/2306802587/PLAT-11535+Slash+commands+arguments
The difficulty here is:
- AntPathMatcher follows the pattern we pass to it. It means we have to rewrite it so that it matches our needs
- A word beginning with @ may not be always an actual mention
- FYI when we check for bot mentions, we currently check for the text content only and not the data.

symphony-elias:
Please read this document before commenting: https://perzoinc.atlassian.net/wiki/spaces/DevX/pages/2306802587/PLAT-11535+Slash+commands+arguments
Thanks

symphony-bdk-core/src/main/java/com/symphony/bdk/core/util/AntPathMatcher.java
symphony-elias: Yes, this is allowed, see : https://www.whitesourcesoftware.com/resources/blog/top-10-apache-license-questions-answered/#3_Is_the_Apache_considered_copyleft

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivity.java
yinan-symphony: @symphony-elias in HFK, i had to read the entity data from the presentationml and match the mention there to the message text to extract the mention name, so far is working well.

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/activity/SlashCommandsWithArgs.java
symphony-elias: Issue here is that we would have two types of consumers:
- either Consumer<CommandContext> without arguments
- or BiConsumer<CommandContext, Map<String, Object>> with arguments
One way could be to accept both, put the arguments map in the context and call whatever handler is provided.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/CommandActivity.java
yinan-symphony: https://github.com/SymphonyOSF/harmony-framework/blob/master/harmony-activities/src/main/java/com/symphony/harmonyframework/activities/command/MentionsReader.java

symphony-bdk-examples/bdk-core-examples/src/main/java/com/symphony/bdk/examples/activity/SlashCommandsWithArgs.java
symphony-youri: Can we put the arguments inside the context object?

symphony-elias:
Implementation done in #600

Issue #598, opened 2021-12-09
Support user suspension API


symphony-mariacristina:
### Description of Problem:
We would like to add support for the Suspend User Account v1 public API in the BDK. 
Link to the API in the documentation: https://developers.symphony.com/restapi/reference#suspend-user-v1

PR #599, opened 2021-12-09
Support user suspension API

symphony-mariacristina:
Closes https://github.com/finos/symphony-bdk-java/issues/598

Adding support for the Suspend User Account v1 public API  (https://developers.symphony.com/restapi/reference#suspend-user-v1)

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
@@ -736,7 +736,6 @@ public class UserService implements OboUserService, OboService<OboUserService> {
    * @param userSuspension action to perform on the user
    * @see <a href="https://developers.symphony.com/restapi/reference#suspend-user-v1">Suspend User Account v1</a>
    */
-  @API(status = API.Status.STABLE)
   public void suspendUser(@Nonnull Long userId, @Nonnull UserSuspension userSuspension) {
     executeAndRetry("suspendUser",
         () -> userApi.v1AdminUserUserIdSuspensionUpdatePut(authSession.getSessionToken(), userId, userSuspension));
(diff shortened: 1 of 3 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -728,6 +729,19 @@ public V1AuditTrailInitiatorList listAuditTrail(@Nonnull Long startTimestamp, Lo
...
   }
 
+  /**
+   * Suspends or re-activates a user account.
+   *
+   * @param userId  user id
+   * @param userSuspension action to perform on the user
+   * @see <a href="https://developers.symphony.com/restapi/reference#suspend-user-v1">Suspend User Account v1</a>
+   */
+  @API(status = API.Status.STABLE)
```
symphony-thibault, line 739: API status is already defined at class level

symphony-thibault (approved):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
symphony-thibault: We could maybe start setting the [`since` param](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html#since()) though :)

symphony-mariacristina: removed! :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -728,6 +729,18 @@ public V1AuditTrailInitiatorList listAuditTrail(@Nonnull Long startTimestamp, Lo
...
   }
 
+  /**
+   * Suspends or re-activates a user account.
+   *
+   * @param userId  user id
+   * @param userSuspension action to perform on the user
+   * @see <a href="https://developers.symphony.com/restapi/reference#suspend-user-v1">Suspend User Account v1</a>
+   */
+  public void suspendUser(@Nonnull Long userId, @Nonnull UserSuspension userSuspension) {
```
symphony-youri, line 739: I was wondering if just "suspend" would be enough but we already have followUser/unfollowUser methods in this class

symphony-youri (approved):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
symphony-mariacristina: At the end we only have one api to suspend, nothing similar. I guess we don't risk any misunderstanding.

PR #600, opened 2021-12-09
#511 Add accessible slash command arguments in the BDK core

symphony-elias:
### Ticket
#511 

### Description
Goal is to be able to define slash commands with arguments that can be later retrieved.
Markdown documentation and an example are missing.

### Checklist


symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/HelpCommand.java
```diff
@@ -20,14 +19,15 @@
  * A help command listing all the commands that can be performed by an end-user through the chat.
  */
 @API(status = API.Status.STABLE)
-public class HelpCommand extends PatternCommandActivity<CommandContext> {
+public class HelpCommand extends SlashCommand {
```
symphony-youri, line 22: don't we risk breaking the compatibility here? (not sure everyone would be extending the HelpCommand, maybe we can check on internal projects)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/Cashtag.java
```diff
@@ -0,0 +1,40 @@
...
+
+import lombok.Getter;
+import org.apiguardian.api.API;
+
+import java.util.Objects;
+
+/**
+ * Class representing a cashtag in a V4Message
+ */
+@API(status = API.Status.INTERNAL)
```
symphony-youri, line 11: maybe a bit of lombok on those simple objects? @Value?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/TypedArgumentToken.java
```diff
@@ -0,0 +1,40 @@
...
+
+  private final Class<T> type;
+  private final String argumentName;
+
+  /**
+   *
+   * @param type the type to be checked for matching
+   * @param pattern the pattern in format "{{@literal @}argName}", "{$argName}" or "{#argName}"
+   * @param <T> the type to be checked for matching
+   * @return
```
symphony-youri, line 20: can be removed

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/parsing/SlashCommandPatternTest.java
```diff
@@ -0,0 +1,314 @@
...
+    return new V4Message().message("<div data-format=\"PresentationML\" data-version=\"2.0\" class=\"wysiwyg\"><p>" + textContent + "</p></div>");
+  }
+
+  private V4Message buildMessage(String textContent, String data) {
+    return buildMessage(textContent).data(data);
+  }
+
+  private void assertIsRegexToken(String expectedRegex, CommandToken actualToken) {
+    assertTrue(actualToken instanceof RegexCommandToken);
+    assertEquals(expectedRegex, ((RegexCommandToken) actualToken).getRegexPattern().pattern());
```
symphony-youri, line 306: in tests I would recommend using assertj for better error messages (and readibility sometime)
not for this change maybe but in the future

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/MatchResult.java
```diff
@@ -0,0 +1,38 @@
...
+ * Class representing the outcome of a matching between a {@link SlashCommandPattern} and a {@link com.symphony.bdk.gen.api.model.V4Message}.
+ * It can contain the map of arguments if applicable. Key is argument name, value is the actual value in the message.
+ * Argument value can be of type {@link String}, {@link Mention}, {@link Cashtag} or {@link Hashtag}.
+ */
+@API(status = API.Status.INTERNAL)
+@Getter
+public class MatchResult {
+
+  private boolean isMatching;
+  private Map<String, Object> arguments;
```
symphony-youri, line 19: would we have an interest in having a custom object here with:
- get(property)
- getMention(property)
- getCashtag(property)
- ....

to save the user from having to cast?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/InputTokenizer.java
```diff
@@ -0,0 +1,179 @@
...
+  @SneakyThrows
+  private static DocumentBuilder initBuilder() {
+    final DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+    return factory.newDocumentBuilder();
+  }
+
+  private Document document;
+  private JsonNode dataNode;
+  private List<Object> tokens;
+  private StringBuffer buffer;
```
symphony-youri, line 59: sonarlint is telling me
_SonarLint: Replace the synchronized class "StringBuffer" by an unsynchronized one such as "StringBuilder"._ 

(I highly suggest using the plugin in Intellij :))

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/command/SlashCommandTest.java
```diff
@@ -67,16 +77,136 @@ void testSlashCommandNotTriggered() {
...
+            + "\"2\":{\"id\":[{\"type\":\"org.symphonyoss.taxonomy.hashtag\",\"value\":\"myhashtag\"}],\"type\":\"org.symphonyoss.taxonomy\"},"
+            + "\"3\":{\"id\":[{\"type\":\"org.symphonyoss.fin.security.id.ticker\",\"value\":\"mycashtag\"}],\"type\":\"org.symphonyoss.fin.security\"}}");
+
+    provider.trigger(l -> l.onMessageSent(new V4Initiator(), event));
+
+    assertTrue(handlerCalled.get());
+    assertEquals(4, arguments.size());
+    assertEquals("value", arguments.get(stringArgName));
+    assertEquals(new Mention("@John Doe", 12345679L), arguments.get(mentionArgName));
+    assertEquals(new Hashtag("#myhashtag", "myhashtag"), arguments.get(hashtagArgName));
```
symphony-youri, line 202: now what it I try to do:
arguments.get(hashtagArgName) and expect to get a string out of it?

for instance if I don't really about the hashtag object but I'm fine to consider it a string

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/MatchResult.java
symphony-youri: Also having our own object would give us immutability

symphony-youri (commented):
mostly minor comments for me (maybe the arguments map or custom object is worth discussing)

now when I see all those changes I wonder how easy is the port to Python going to be :)

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/command/HelpCommand.java
symphony-elias: Yes indeed. I discussed it with @symphony-thibault 
We assumed customers would either use SlashCommands through the static methods or or directly subclass PatterCommandActivity

symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/command/SlashCommandTest.java
symphony-elias: We could have several methods like `getAsString(argName)` that would always return a string representation and specific methods like: `getMention(argName)` etc.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/Cashtag.java
```diff
@@ -0,0 +1,21 @@
+package com.symphony.bdk.core.activity.parsing;
+
+import lombok.Value;
+import org.apiguardian.api.API;
+
+/**
+ * Class representing a cashtag in a V4Message
+ */
+@API(status = API.Status.INTERNAL)
```
symphony-thibault, line 9: Are they really `INTERNAL`? I think we expose them to the developers, let's make them `STABLE` instead

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/TypedArgumentToken.java
symphony-elias: I'd prefer to keep the comments

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
```diff
@@ -72,10 +72,11 @@ private void preProcessActivity(AbstractActivity<?, ?> activity) {
           abstractActivity.getInfo().name());
     });
 
-    // a command activity (potentially) needs the bot display name in order to parse the message text content
+    // a command activity (potentially) needs the bot display name and user ID in order to parse the message text content
     // this way of passing this information is not very clean though, we should find something
     if (activity instanceof CommandActivity) {
       ((CommandActivity<?>) activity).setBotDisplayName(this.botSession.getDisplayName());
```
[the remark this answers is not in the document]
symphony-elias: This is used in the CommandActivity and exposed with a public getter. So it means we will break compatibility.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/TypedArgumentToken.java
symphony-youri: I meant the empty @return tag

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/parsing/Arguments.java
```diff
@@ -0,0 +1,82 @@
...
+    final Object argumentValue = arguments.get(argumentName);
+    return argumentValue == null ? null : argumentValue.toString();
+  }
+
+  /**
+   *
+   * @param argumentName the name of the argument to be retrieved
+   * @return the argument value if present and if is of type {@link Hashtag}, null otherwise
+   */
+  public Hashtag getAsHashtag(String argumentName) {
```
symphony-thibault, line 52: I'd prefer `getHashtag(String argumentName)` instead of `getAsHashtag(String argumentName)`. WDYT @symphony-elias @symphony-youri ?

symphony-thibault: Yes I think we would have a `getString` as well, and potentially `getBoolean` or `getInteger` in the future

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/ActivityRegistry.java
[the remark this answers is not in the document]
symphony-thibault: Correct, let's keep it for backward compatibility 👍

symphony-youri (approved):

PR #601, opened 2021-12-10
Support user suspension API

symphony-mariacristina:
Related to https://github.com/finos/symphony-bdk-java/issues/598, we decided to ease the usage of this endpoint by providing two different functions to suspend/unsuspend a user account.
- suspendUser(userId, reason, until)
- unsuspendUser(userId)

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
@@ -730,13 +730,33 @@ public class UserService implements OboUserService, OboService<OboUserService> {
   }
 
   /**
-   * Suspends or re-activates a user account.
+   * Suspends a user account.
+   * Calling this endpoint requires a service account with the User Provisioning role.
    *
-   * @param userId  user id
-   * @param userSuspension action to perform on the user
+   * @param userId  user id to suspend
+   * @param reason  reason why the user has to be suspended
+   * @param until   timestamp in milliseconds
    * @see <a href="https://developers.symphony.com/restapi/reference#suspend-user-v1">Suspend User Account v1</a>
    */
-  public void suspendUser(@Nonnull Long userId, @Nonnull UserSuspension userSuspension) {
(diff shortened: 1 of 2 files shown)


symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/user/UserService.java
```diff
@@ -730,13 +730,33 @@ public V1AuditTrailInitiatorList listAuditTrail(@Nonnull Long startTimestamp, Lo
...
    *
-   * @param userId  user id
-   * @param userSuspension action to perform on the user
+   * @param userId  user id to suspend
+   * @param reason  reason why the user has to be suspended
+   * @param until   timestamp in milliseconds
    * @see <a href="https://developers.symphony.com/restapi/reference#suspend-user-v1">Suspend User Account v1</a>
    */
-  public void suspendUser(@Nonnull Long userId, @Nonnull UserSuspension userSuspension) {
+  public void suspendUser(@Nonnull Long userId, @Nonnull String reason, @Nonnull Long until) {
```
symphony-thibault, line 741: That could be more developer friendly to pass a `LocalDateTime` instead of a `Long` value here. WDYT?

symphony-youri: I would maybe not check the validity as it is server side logic

symphony-youri: maybe just an Instant instead of Long or LocalDateTime?

symphony-mariacristina: I checked in the other endpoints we are using Instant for example for the "since" query parameter. IMO I would go for that in here as well.

symphony-mariacristina: Done https://github.com/finos/symphony-bdk-java/pull/601/commits/f172a527fd55cf4b29bc52224cc3f757026e90a1

symphony-youri (approved):

symphony-elias (approved):
LGTM

PR #603, opened 2021-12-14
#511 Make slash command arguments available in SB starter

symphony-elias:
### Ticket
#511

### Description
Make slash command arguments available in annotated methods with @Slash in SB starter

### Checklist

### DIFF
--- a/docs/spring-boot/core-starter.md
+++ b/docs/spring-boot/core-starter.md
@@ -291,6 +291,49 @@ public class SlashHello {
 By default, the `@Slash` annotation is configured to require bot mention in order to trigger the command. You can override
 this value using `@Slash#mentionBot` annotation parameter. 
 
+You can also use slash commands with arguments. To do so, the field `value` of the `@Slash` annotation must have a valid
+format as explained in the [Activity API section](../activity-api.md#Slash-command-pattern-format). 
+If the slash command pattern is valid, you will have to specify all slash arguments as method parameter with the same name and type.
+If slash command pattern or method signature is incorrect, a `warn` message will appear in your application log and
+the slash command will not be registered.
+
+For instance:
+```java
+@Component
+public class SlashHello {
+
+  @Slash("/hello {arg") // will not be registered: invalid pattern
+  public void onHelloInvalidPattern(CommandContext commandContext, String arg) {
+    log.info("On /hello command");
+  }
+
+  @Slash("/hello {arg1}{arg2}") // will not be registered: invalid pattern
+  public void onHelloInvalidPatternTwoArgs(CommandContext commandContext, String arg1, String arg2) {
+    log.info("On /hello command");
+  }
+
+  @Slash("/hello {arg1} {arg2}") // will be registered: valid pattern and valid signature
+  public void onHelloValidPatternTwoArgs(CommandContext commandContext, String arg1, String arg2) {
+    log.info("On /hello command");
+  }
+
+  @Slash("/hello {arg1} {arg2}") // will not be registered: valid pattern but missing argument
+  public void onHelloValidPatternTwoArgs(CommandContext commandContext, String arg1) {
+    log.info("On /hello command");
+  }
+
+  @Slash("/hello {arg1} {@arg2}") // will not be registered: valid pattern but mismatching type for arg2
+  public void onHelloValidPatternTwoArgs(CommandContext commandContext, String arg1, String arg2) {
+    log.info("On /hello command");
+  }
+
+  @Slash("/hello {arg1} {@arg2} {#arg3} {$arg4}") // will be registered: valid pattern and correct signature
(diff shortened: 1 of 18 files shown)


symphony-elias:
In this implementation, methods annotated `@Slash` with a command pattern containing arguments must have all arguments declared as method parameters in order to be registered and triggered. Is it something we want to enforce (since everything is in the `CommandContext`)?

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
```diff
@@ -32,6 +37,8 @@
  * @see <a href="https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/context/event/EventListenerMethodProcessor.java">EventListenerMethodProcessor.java</a>
  */
 @Slf4j
+@Generated // means excluded from test coverage: error cases hard to test,
```
symphony-thibault, line 40: Could we move this annotation on specific methods that we can't fully cover? Assuming that most of the class can still be correctly covered

symphony-thibault (approved):
LGTM 👍

symphony-youri (commented):
Does an existing command like "/command blah" still works?

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
```diff
@@ -43,6 +48,13 @@
       "com.symphony.bdk.core."
   };
 
+  private static final ParameterNameDiscoverer PARAMETER_NAME_DISCOVERER = new DefaultParameterNameDiscoverer();
+
+  /**
+   * Key is the annotated slash method, value is the map of (parameter name, index)
+   */
+  private static final Map<Method, Map<String, Integer>> METHOD_TO_ARGUMENT_INDEXES = new HashMap<>();
```
symphony-youri, line 56: do we use it as a cache? I would expect to find in the code something where we try to read an entry and if we don't find it we initialize, here it looks like it is used as a regular field (or maybe I missed something)

docs/spring-boot/core-starter.md
```diff
@@ -291,6 +291,49 @@ public class SlashHello {
 By default, the `@Slash` annotation is configured to require bot mention in order to trigger the command. You can override
 this value using `@Slash#mentionBot` annotation parameter. 
 
+You can also use slash commands with arguments. To do so, the field `value` of the `@Slash` annotation must have a valid
+format as explained in the [Activity API section](../activity-api.md#Slash-command-pattern-format). 
+If the slash command pattern is valid, you will have to specify all slash arguments as method parameter with the same name and type.
+If slash command pattern or method signature is incorrect, a `warn` message will appear in your application log and
```
symphony-youri, line 297: or would it be better to fail the startup? at least the user is aware he has made a mistake

symphony-youri: mm the warning was there before... would it be a breaking change to change that....

symphony-elias: Yes, I would say it is a breaking change. Let's discuss that with @symphony-thibault

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
symphony-elias: Yes and it is initialized line 215. I will update the implementation to make it clearer.

symphony-elias:
> Does an existing command like "/command blah" still works?

Yes, it was tested here: symphony-bdk-core/src/test/java/com/symphony/bdk/core/activity/parsing/SlashCommandPatternTest.java
I will more tests in SlashCommandTest to make it more obvious.

symphony-bdk-spring/symphony-bdk-core-spring-boot-starter/src/main/java/com/symphony/bdk/spring/annotation/SlashAnnotationProcessor.java
symphony-elias: Well, it seems we don't really need a cache. The DefaultNameParameterNameDiscoverer already caches some of the results (see LocalVariableTableParameterNameDiscoverer in SB)

docs/spring-boot/core-starter.md
symphony-thibault: @symphony-elias @symphony-youri I'd keep the warning for now. We could maybe propose that as a possible improvement and let the community decide (through a GH issue)

Issue #479, opened 2021-03-28
Module system 


symphony-thibault:
> **DRAFT, IN PROGRESS**

The purpose of this issue is to introduce and specify a new module system for the BDK library. We want to empower contributors to create their own modules or plugins in separate repositories, without necessary having to make their contributions directly in the BDK codebase. 

## Overview

```java
public class MyModule implements BdkModule, BdkLifecycleAware, BdkConfigurationAware {
  
  @Override // provided by BdkConfigurationAware
  public void onConfigurationSet(BdkConfig config) {}

  @Override // provided by BdkLifecycleAware
  public void onCoreInitialized(SymphonyBdk bdk) {}
}
```

## Loading system
To explore: 
- manual loading through [`SymphonyBdkBuilder`](https://javadoc.io/doc/org.finos.symphony.bdk/symphony-bdk-core/latest/com/symphony/bdk/core/SymphonyBdkBuilder.html): 
```java
SymphonyBdk bdk = new SymphonyBdkBuilder().config(config).withModule(MyModule.class).build();
```
- using JDK [`ServiceLoader`](https://docs.oracle.com/javase/9/docs/api/java/util/ServiceLoader.html)

## Module Service
On the other hand, a developer should also be able to retrieve a module service, for instance using: 

```java
MyModuleService moduleService = bdk.lookupModuleService(MyModuleService.class); 
```

symphony-youri:
Supporting an Agentless architecture in the BDK is one of the use case for extension.

While the service loader approach has the benefit of being natively supported it is also less explicit and a bit of magic happening at runtime.
The module approach with clear and well defined extension points might help to make it more explicit. We could take inspiration from Jackson modules for instance.

For instance, here is an extension point that was defined:

```java
public interface BdkModule {

  <T> T getService(AuthSession authSession, ApiClientFactory apiClientFactory, RetryWithRecoveryBuilder<?> retryBuilder,
      Class<T> messageServiceClass);

}
```

What we want to make sure is to encapsulate the method's parameters in an object to easily evolve those without breaking compatibility (at least we would be free to pass new data around).

PR #613, opened 2021-12-24
PoC of extensibility mechanism

symphony-elias:
### Ticket
#479 

### Description
PoC demonstrating what could be an extensibility mechanism in the BDK (core and Spring Boot starter).

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
@@ -10,6 +10,10 @@ import com.symphony.bdk.core.auth.exception.AuthUnauthorizedException;
 import com.symphony.bdk.core.client.ApiClientFactory;
 import com.symphony.bdk.core.config.exception.BotNotConfiguredException;
 import com.symphony.bdk.core.config.model.BdkConfig;
+import com.symphony.bdk.core.extension.AuthSessionAware;
+import com.symphony.bdk.core.extension.BdkConfigAware;
+import com.symphony.bdk.core.extension.Extension;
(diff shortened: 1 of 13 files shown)


symphony-bdk-examples/bdk-spring-boot-example/src/main/java/com/symphony/bdk/examples/spring/EchoActivity.java
```diff
@@ -21,6 +22,9 @@
   @Autowired
   private MessageService messageService;
 
+  @Autowired
+  private MyExtension extension;
```
symphony-youri, line 26: Cool

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
```diff
@@ -42,6 +49,7 @@
 public class SymphonyBdk {
 
   private final BdkConfig config;
+  private final List<Extension> extensions;
```
symphony-youri, line 52: what is the value of storing over instanciating them on the fly in the getExtension() method? 
I would say to keep the same lifecycle as other services (but just to make sure and to open the discussion)

symphony-youri (commented):
Maybe we should name them ServiceExtension? (a bit long...) to avoid blocking the concept of extension in case we come up with other extension points?

symphony-bdk-core/src/main/java/com/symphony/bdk/core/SymphonyBdk.java
symphony-elias: Indeed, the added value is low and we could instantiate them on the fly.

PR #622, opened 2022-01-11
#621 Added streamId and formMessageId to FormReplyActivity

### LINKED ISSUE #621: Add Message ID and Stream ID to FormReplyActivity's FormReplyContext
### Description of Problem:
Using FormReplyActivity to capture Elements form submissions currently does not provide the Message ID from the originating form. This blocks workflows attempting to either suppress or update the message containing the form that was just submitted.

For message update, message id alone is not sufficient and stream id is also required. Obtaining stream id currently requires a call 3 levels deep: `context.getSourceEvent().getStream().getStreamId()`.

### Potential Solutions:
Add `getMessageId()` and `getStreamId()` to FormReplyContext. This will allow the most commonly-used identifiers - Form ID, Message ID and Stream ID - to be accessed easily when processing form submissions.

symphony-thibault:
Closes #621

2 new fields added to the `FormReplyActivity` class: 
- `streamId`
- `formMessageId`

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyContext.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyContext.java
@@ -19,6 +19,12 @@ import javax.annotation.Nullable;
 @API(status = API.Status.STABLE)
 public class FormReplyContext extends ActivityContext<V4SymphonyElementsAction> {
 
+  /** The messageId extracted from event source */
+  private String formMessageId;
+
+  /** The messageId extracted from event source */
+  private String streamId;
+
   /** The formId extracted from event source */
   private String formId;
 
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyContext.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyContext.java
@@ -19,10 +19,10 @@ import javax.annotation.Nullable;
 @API(status = API.Status.STABLE)
 public class FormReplyContext extends ActivityContext<V4SymphonyElementsAction> {
 
-  /** The messageId extracted from event source */
+  /** The "form" messageId (different from the "reply" messageId) extracted from event source */
   private String formMessageId;
 
-  /** The messageId extracted from event source */
+  /** The streamId extracted from event source */
   private String streamId;
 
   /** The formId extracted from event source */
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyActivity.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyActivity.java
@@ -33,7 +33,11 @@ public abstract class FormReplyActivity<C extends FormReplyContext>
   @Override
   protected void beforeMatcher(C context) {
     super.beforeMatcher(context);
-    // copy formId at context root level
+    // copy streamId to context root level
+    context.setStreamId(context.getSourceEvent().getStream().getStreamId());
+    // copy formMessageId to context root level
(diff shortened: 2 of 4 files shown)


symphony-mariacristina (approved):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyContext.java
```diff
@@ -19,6 +19,12 @@
 @API(status = API.Status.STABLE)
 public class FormReplyContext extends ActivityContext<V4SymphonyElementsAction> {
 
+  /** The messageId extracted from event source */
+  private String formMessageId;
+
+  /** The messageId extracted from event source */
```
symphony-soufiane, line 25: streamId not messageId in comment

symphony-soufiane (approved):

symphony-bdk-core/src/main/java/com/symphony/bdk/core/activity/form/FormReplyContext.java
symphony-thibault: Fixed via e3ccd3e

symphony-youri (approved):

ystan-:
@symphony-thibault Sorry there seems to be some confusion here: `formMessageId` isn't the same as `messageId`. To suppress or update the message containing the form that triggered the `SYMPHONYELEMENTSACTION` event, we will need the top-level `messageId` rather than `symphonyElementsAction.formMessageId`.

https://docs.developers.symphony.com/building-bots-on-symphony/datafeed/real-time-events#symphony-elements-action

symphony-thibault:
Hi @ystan-, 

To me what has been done in this PR is correct and resolve the problem: 
> Using FormReplyActivity to capture Elements form submissions currently does not provide the Message ID from the originating form. This blocks workflows attempting to either suppress or update the message containing the form that was just submitted.

The top-level `messageId` from the [symphony-elements-action](https://docs.developers.symphony.com/building-bots-on-symphony/datafeed/real-time-events#symphony-elements-action) payload corresponds to the _formReply_ messageId (that is hidden in the chat).  

The following example shows that what is needed to update the original form is the `formMesssageId` field and not the _formReply_ messageId: 
```java
@Slf4j
public class Example {

  private static final String FORM = "<messageML>\n"
      + "  <form id=\"the-form\">\n"
      + "    <text-field name=\"name\" placeholder=\"Input your name...\" required=\"true\"/>\n"
      + "    <button name=\"send-answers\" type=\"action\">Send Answers</button>\n"
      + "  </form>\n"
      + "</messageML>";

  public static void main(String[] args) throws Exception {

    final SymphonyBdk bdk = new SymphonyBdk(loadFromSymphonyDir("config.yaml"));

    bdk.activities().register(slash("/form", false, c -> {
      V4Message msg = bdk.messages().send(c.getStreamId(), FORM);
      log.info("Message sent: {}", msg.getMessageId());
    }));

    bdk.datafeed().subscribe(new RealTimeEventListener() {

      @Override
      public void onSymphonyElementsAction(V4Initiator initiator, V4SymphonyElementsAction event) throws EventException {
        log.info("Form reply received: {}", event.getFormMessageId());
      }
    });

    bdk.datafeed().start();
  }
}
``` 
with output: 
```
16:00:21.015 [main] Vxs9UV INFO  c.s.bdk.examples.Example - Message sent: Tud9Lauq2CeBSLDzqXVWRn___oG2WIGcbQ
16:00:55.691 [main] Lwd6zZ INFO  c.s.bdk.examples.Example - Form reply received: Tud9Lauq2CeBSLDzqXVWRn___oG2WIGcbQ
```
We indeed decided to call this field `formMessageId` to avoid any confusion with the _formReply_ messageId. 

Please let me know if it helps 👍

ystan-:
Apologies - you're right. This works.

PR #633, opened 2022-03-10
Supported a list of files in multipart form data

### LINKED ISSUE #631: Limitation for multipart/form-data form
## Feature Request
The current httpClient of bdk doesn't support multiple values for a given form key. 
With the current httpClient of bdk, this works: 
```java
httpClient.formParam("file", new File("path/to/myFile/");
```

But when we keep adding other files like the snippet bellow doesn't work because we keep overriding the map containing the parameters:
```java
httpClient.formParam("file", new File("path/to/myFirstFile/");
httpClient.formParam("file", new File("path/to/mySecondFile/");
```

### Description of Problem:
 I checked the code and it seems like the form params are stored in a map, and the value of "file" will be overridden each time which will prevent us from having a field with many files. The implementation ideally would support adding of multiple files under the same key and use boundary to separate the files under the field **_file_**.

Ideally we want to to be able to add several files in one field of form data in an iterative way, like this:
```java
httpClient.formParam("file", new File("path/to/myFirstFile/");
httpClient.formParam("file", new File("path/to/mySecondFile/");
httpClient.formParam("file", new File("path/to/myThirdFile/");
```

### Potential Solutions:
N/A

symphony-elias:
### Description
Closes #631

### Checklist
- [x] Referenced an issue in the PR title or description
- [x] Unit/Integration tests updated or added
- [N/A] Javadoc added or updated
- [N/A] Updated the documentation in [docs folder](../docs)

### DIFF
--- a/symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/HttpClient.java
+++ b/symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/HttpClient.java
@@ -10,6 +10,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.With;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -375,13 +376,17 @@ public class HttpClient {
 
     private Map<String, String> headers;
     private Map<String, String> cookies;
-    private Map<String, Object> formParams;
+    private Map<String, List<Object>> formParams;
     private List<Pair> queryParams;
 
     private String path;
     private Object body;
     private String accept;
     private String contentType;
+
+    public Map<String, Object> getFormParams() {
+      return formParams == null ? Collections.emptyMap() : new HashMap<>(formParams);
+    }
 
     public Map<String, String> appendHeader(String key, String value) {
 
@@ -413,13 +418,16 @@ public class HttpClient {
       return this.queryParams;
     }
 
-    public Map<String, Object> appendFormParam(String key, Object value) {
+    public Map<String, List<Object>> appendFormParam(String key, Object value) {
 
       if (this.formParams == null) {
         this.formParams = new HashMap<>();
       }
+      if (!this.formParams.containsKey(key)) {
+        this.formParams.put(key, new ArrayList<>());
+      }
 
-      this.formParams.put(key, value);
+      this.formParams.get(key).add(value);
       return this.formParams;
     }
   }
--- a/symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/util/ApiUtils.java
+++ b/symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/util/ApiUtils.java
@@ -4,15 +4,17 @@ import org.apiguardian.api.API;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.TrustManagerFactory;
-import javax.net.ssl.X509TrustManager;
-
+import java.io.File;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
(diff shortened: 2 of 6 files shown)


symphony-bdk-http/symphony-bdk-http-webclient/src/main/java/com/symphony/bdk/http/webclient/ApiClientWebClient.java
```diff
@@ -246,6 +249,11 @@ private void serializeMultiPartDataEntry(String paramKey, Object paramValue,
     }
   }
 
+  private void addFileToFormParam(String paramKey, File paramValue, MultiValueMap<String, Object> formValueMap) {
+    File file = paramValue;
+    formValueMap.add(paramKey, new FileSystemResource(file));
```
imadamahmid1, line 254: does this file get stored in the machine?

symphony-elias: The file has to be stored on the machine and accessible by the Java process running this. 
I guess your question is: does the file get closed after HTTP request is made. I have to check that.

symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/HttpClient.java
```diff
@@ -413,13 +417,16 @@ public HttpClient build() {
       return this.queryParams;
     }
 
-    public Map<String, Object> appendFormParam(String key, Object value) {
+    public Map<String, List<Object>> appendFormParam(String key, Object value) {
```
symphony-youri, line 420: aren't we breaking the API if we change that?
but then it looks like this part is internal only

symphony-youri (commented):
🙏 put a description in the commit message to give a bit of context

symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/util/ApiUtils.java
```diff
@@ -52,6 +54,15 @@ public static void addDefaultRootCaCertificates(KeyStore trustStore) throws Gene
     }
   }
 
+  public static boolean isCollectionOfFiles(Object paramValue) {
+    if (!(paramValue instanceof Collection<?>)) {
+      return false;
+    }
+
+    final Collection<?> collection = (Collection<?>) paramValue;
+    return !collection.isEmpty() && collection.stream().findAny().get() instanceof File;
```
symphony-youri, line 63: should we check that all elements are files?

symphony-bdk-http/symphony-bdk-http-api/src/main/java/com/symphony/bdk/http/api/HttpClient.java
symphony-elias: Yes this is internal only

symphony-bdk-http/symphony-bdk-http-jersey2/src/main/java/com/symphony/bdk/http/jersey2/ApiClientJersey2.java
```diff
@@ -430,18 +432,12 @@ protected boolean isJsonMime(String mime) {
...
-            .size(file.length())
-            .build();
-        final FormDataBodyPart streamPart = new FormDataBodyPart(
-            contentDisposition,
-            file,
-            MediaType.APPLICATION_OCTET_STREAM_TYPE
-        );
-        multiPart = (FormDataMultiPart) multiPart.bodyPart(streamPart);
+        multiPart = buildFileBodyPart(multiPart, param.getKey(), (File) param.getValue());
+      } else if (isCollectionOfFiles(param.getValue())) {
```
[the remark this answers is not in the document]
symphony-elias: If it is a collection of something else, we will go to the else statement:
`multiPart = multiPart.field(param.getKey(), this.parameterToString(param.getValue()));`

symphony-youri (approved):

symphony-mariacristina (approved):

Issue #510, opened 2021-05-03
Some links in the GH pages documentation are dead


symphony-elias:
## Bug Report

In https://symphony-bdk-java.finos.org/datafeed.html (or https://github.com/finos/symphony-bdk-java/blob/main/docs/activity-api.md) most of the external links are invalid following the revamp of the Symphony developers documentation. Links all redirect to the home page of the Symphony developers documentation.
https://symphony-bdk-java.finos.org/symphony-bdk-examples/bdk-multi-instances-example leads to 404 error.

## Expected result
External links in the documentation should be up to date

RanabirChakraborty:
Hello @symphony-elias can I take a look into it?

symphony-elias:
> Hello @symphony-elias can I take a look into it?

Hello @RanabirChakraborty of course! Feel free to open a PR referencing this issue

PR #644, opened 2022-03-28
Fixing broken documentation links

symphony-mariacristina:
### Description
Some of the documentation links were outdated, goal of this implementation is to replaced them with working links.

Closes https://github.com/finos/symphony-bdk-java/issues/510


symphony-youri (approved):
Wow that was a lot of links
(breaking such URLs is not good for referencing :()

Will you do the same for the Python BDK?

symphony-mariacristina:
Yes @symphony-youri , I'm going to do the same for Python.

PR #672, opened 2022-08-11
#670: Retry error message does not use the correct api base path in AbstractExtensionAppAuthenticator

### LINKED ISSUE #670: Retry error message does not use the correct api base path in AbstractExtensionAppAuthenticator
## Bug Report
When the request to get pod certificate fails and gets retried, the error message being built contains wrong base path which is confusing:

https://github.com/finos/symphony-bdk-java/blob/main/symphony-bdk-core/src/main/java/com/symphony/bdk/core/auth/impl/AbstractExtensionAppAuthenticator.java#L58

getBasePath() method does always return base path for the login api client while we need the pod api client one in this case.

symphony-soufiane:
### Description
Closes #670 
ExtensionAppAuthenticatorRsaImpl class has 2 different api clients: login and pod api clients with different base paths. The overridden method getBasePath() always returns login pod api base path. 
When the api call to get pod certificate fails and gets retried, the getBasePath() is called to provide more info in the error message. In this case, the login pod api client base path is injected while the pod api client is the one being used. 

In this PR we separate both apis base path with 2 different methods to avoid confusion: 

Added two methods, one to get the basepath used to get pod certificate and another one to get the basepath used to authenticate and retrieve tokens.


ystxn (approved):
LGTM

PR #737, opened 2022-12-08
Add actuator health endpoint to app starter

yinan-symphony:
The health endpoint uses the HealthService to provide the core components health status, as well as the DatafeedLoop running process status, if the running process counter is not updated in a fixed time range, the process is considered as 'DOWN'.

### Description
Closes #[ISSUE NUMBER]

Please put here the intent of your pull request.

### Dependencies

### Checklist
- [ ] Referenced an issue in the PR title or description
- [ ] Unit/Integration tests updated or added

### DIFF
--- a/docs/spring-boot/app-starter.md
+++ b/docs/spring-boot/app-starter.md
@@ -4,6 +4,7 @@ The Symphony BDK for Java provides a _Starter_ module that aims to ease extensio
  
 ## Features
 - Configure extension app through `application.yaml`
+- Application health info through Spring Boot standard actuator endpoint
 - Expose a REST API to perform [Circle of Trust authentication](https://docs.developers.symphony.com/building-extension-applications-on-symphony/app-authentication/circle-of-trust-authentication).
 
 ## Installation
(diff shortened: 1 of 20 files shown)


docs/spring-boot/app-starter.md
```diff
@@ -126,3 +127,23 @@ public class ExtAppSpringApplication {
...
+
+```
+management:
+  endpoint:
+    web:
+      exposure:
+        include: 'symphonyBdk'
+    health:
+      show-details: "ALWAYS"
+```
```
symphony-soufiane, line 149: we need maybe to update configuration.md too with this new added config: https://github.com/finos/symphony-bdk-java/blob/main/docs/configuration.md

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedLoop.java
```diff
@@ -37,4 +37,11 @@ public interface DatafeedLoop {
      * @param listener a Datafeed event listener to be unsubscribed
      */
     void unsubscribe(RealTimeEventListener listener);
+
+  /**
+   * The timestamp of the last successful pulling
+   *
+   * @return timestamp value in long
+   */
+  long lastPullTimestamp();
```
symphony-soufiane, line 46: I don't think we want to expose this method in DF loop api to users.

symphony-soufiane (approved):
few comments.

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/datafeed/DatafeedLoop.java
yinan-symphony: then how to get this timestamp in a generic way ? another way is to move the health check method inside the DatafeedLoop,  but i would say expose a last pull timestamp method in this interface is more relevant than exposing a health check method.

docs/spring-boot/app-starter.md
yinan-symphony: Just check this configuration.md content, il looks like it s more like BDK configuration only, if we put this into configuration.md, i feel that it s kinda mix between BDK and spring boot config, especially considering that this config is in app-starter only.

Issue #680, opened 2022-10-12
BDK ability to turn on and off the body tag payload. For datafeed read and create calls.


ming-kwong:
### Description of Problem:
Customer encountered an issue with their bot on the 24-Sept, where their BOT unable to create any datefeed.
There was an issue datafeed service where there was some orphan value left in the dynamoDB. Preventing datafeed to be created/used.
However when they create the datafeed without the body tag payload, the datafeed works fine.
Customer had to rollback to datafeed v1 completely and wait for us to remove the orphan value in the dynamoDB

### Potential Solutions:
Currently BDK includes that body tag payload as a default.
A flag in the config to toggle on/off the body payload would be useful. So customer doesnt need to switch to datafeed v1 completely.

PR #738, opened 2022-12-08
#680: not set bot username as Datafeed tag

symphony-soufiane:
### Description
The Datafeed tag is an internal feature and the tag property is optional in the configuration. So far, when no tag is configured, the bot username was used as a tag. This leads to production issues when the tag is an old hanging one. In this PR, we only set the tag in Datafeed calls when it is provided in configuration, otherwise a null value is put. For the same occasion, I corrected the Datafeed documentation with some links returning a 404 page error. The tag property is not documented as it is only internal feature.

### DIFF
--- a/docs/datafeed.md
+++ b/docs/datafeed.md
@@ -8,9 +8,9 @@
 The datafeed loop is a service used for handling the [_Real Time
 Events_](https://docs.developers.symphony.com/building-bots-on-symphony/datafeed/real-time-events). When a user makes an interaction within the IM,
 MIM or Room chat like sending a message, joining or leaving a room chat..., when a connection request is sent, when a
-wall post is published or when a user replies an Symphony element, an event will be sent to the datafeed. The bot can
+wall post is published or when a user replies to a Symphony element, an event will be sent to the datafeed. The bot can
 create a datafeed, list all created datafeeds or retrieve all the Real Time Events within a datafeed through datafeed
-API. The datafeed loop is a core service built on top of the Datafeed API and provide a dedicated contract to bot
+API. The datafeed loop is a core service built on top of the Datafeed API and provides a dedicated contract to bot
 developers to work with datafeed.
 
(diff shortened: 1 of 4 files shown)


docs/datafeed.md
```diff
@@ -62,11 +62,10 @@ datafeed:
...
 
-The minimal configuration for the datafeed service is the version of the datafeed which will be chosen to be use in the
-BDK. For the moment, not all the customers have the datafeed version 2 available on their systems, that's why bot
-developers are able to choose the datafeed version that they wish to use on their bot. If the bot developers want to use
-the datafeed version 2 in their bot, the version configuration have to be specified as `v2`. Otherwise, the datafeed
-version 1 will be used by default.
+The minimal configuration for the datafeed service is the version to be used in the
+BDK. For the moment, not all customers have datafeed 2 available on their systems, that's why bot
+developers are able to choose the version that they wish to use on their bot. If the bot developers want to use
+datafeed 2, the version configuration have to be specified as `v2`. Otherwise, datafeed 1 will be used by default.
```
yinan-symphony, line 68: I thought v2 had been set as default one, is not the case?

symphony-bdk-config/src/main/java/com/symphony/bdk/core/config/model/BdkDatafeedConfig.java
```diff
@@ -15,6 +15,7 @@ public class BdkDatafeedConfig {
 
   private String version = "v2";
```
yinan-symphony, line 16: actually, yes, v2 is the default value now, the doc must be adapt accordingly.

docs/datafeed.md
symphony-soufiane: Yes DFv2 is the default. I was just polishing the writing but the doc itself was outdated. Thanks for raising that, I corrected it in the last commit!

yinan-symphony (approved):

Issue #741, opened 2023-02-06
Missing accessors to get the Datafeed Event Timestamp/initiator


ming-kwong:
### Description of Problem:
Timestamp of datafeed events are missing, users are use the datafeed read timestamp as the timestamp in most cases, but if bot/app reconnect after a disconnection. The actual timestamp of the datafeed events are lost as there are no accessors to get them. The datafeed read timestamp, would be different to when the event actually happen.

Example: Only the stream and affectedUser is available. Its missing the initiator and the event timestamp.
`2023-02-03 16:18:17.745 DEBUG 2704 --- [onyBdk_Datafeed] c.s.b.c.s.d.impl.AbstractDatafeedLoop    : Before dispatching 'USERLEFTROOM' event to listener com.symphony.new_bot.OnUserJoinedRoomListener@1805383
event: class V4UserLeftRoom {
    stream: class V4Stream {
        streamId: PaA58pDQWclZep1oIc12Ln___nnqV15ydA
        streamType: ROOM
        roomName: ttttessstttt
        members: null
        external: false
        crossPod: false
    }
    affectedUser: class V4User {
        userId: 344147139497768
        firstName: null
        lastName: null
        displayName: MingBot2
        email: MingBot2@sup.symphony.com
        username: MingBot2
    }
}`

### Potential Solutions:
Add back getTimestamp for all datafeed events

ming-kwong:
raising on behalf of customer for support case 89156

PR #748, opened 2023-03-16
Fix the event timestamp value is not accessible bug

yinan-symphony:
The event timestamp is filled at the event wrapper level only, which is the one when the event is read from datafeed. However only the raw event is pushed to the real time listeners, that's the reason why we loose this event timestamp.

This commit fix this issue by constructing an event proxy at runtime, all events implements EventPayload interface, the timestamp value is now accessible through this interface. User has to cast the raw event to EventPayload to read the value. This big advantage is we do not bring in any breaking changes in any stable APIs, the change is totally transparent for the BDK developers.

### Description
Closes #[741](https://github.com/finos/symphony-bdk-java/issues/741)

### Dependencies

### Checklist
- [ ] Referenced an issue in the PR title or description
- [ ] Unit/Integration tests updated or added

### DIFF
--- a/allow-list.xml
+++ b/allow-list.xml
@@ -67,4 +67,11 @@
       <gav>org.json:json:20220320</gav>
       <cve>CVE-2022-45688</cve>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+          No fix available
+          ]]></notes>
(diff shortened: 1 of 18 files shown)


ystxn (approved):
LGTM

PR #757, opened 2023-04-07
Migrate bdk to use Springboot 3

symphony-soufiane:
### Description
In order to fix CVE-2023-20861, bdk needed a migration to use Springboot 3. This latter requires java 17 as runtime which leads to some changes in the code. The most important changes are the use of jakarta.ws.rs instead of javax library that is not provided by the java 17 core. Security filter has changed in SecurityConfig class. Java 17 does not allow to make private field as public in UnmodifiableMap system environment reflection which breaks the unit testBdkConfigParserTest. This later still needs a fix to be done in next commit


yinan-symphony (commented):
the github workflow actions have to be changed to jdk17, too.

PR #797, opened 2024-03-05
Add update message in OboMessageService

antoinerollindev:
This commit updates the OboMessageService so that a BDK bot can update a message on behalf of someone (along with OBO message send and OBO message delete).


yinan-symphony (approved):
LGTM

PR #853, opened 2026-02-03
CAIP-302 Add includeInvisible agent datafeed parameter

symphony-enrico:
### Description
By default, the agent filters out invisible room events from its data feeds. Add a configuration parameter to allow these events to be included (false by default).

Added also an optional configuration parameter to specify a datafeed tag to pick


FabienVSymphony (approved):

thibauult (approved):
LGTM 👍🏻

PR #858, opened 2026-02-09
Improve managing <messageML> tag

symphony-enrico:
### Description
In previous versions of the BDK, the wrapping check was less restrictive. The logic has recently been updated to a stricter implementation.

Previous
```java
  if (!this.content.startsWith("<messageML>") && !this.content.endsWith("</messageML>")) {
```

Current:
```java
   boolean isAlreadyWrapped = this.content.startsWith("<messageML") && this.content.endsWith("</messageML>");
```

The Issue: This change broke compatibility with existing template files that include trailing characters, such as a Carriage Return (CR) or newlines after the closing </messageML> tag. While technically a template formatting issue, the current check is too rigid for real-world production environments where IDEs or OS-specific line endings might be appended.

To restore backward compatibility and improve robustness, I have implemented a more tolerant check that ignores leading and trailing whitespace, while throwing also exception if only one tag is present

### DIFF
--- a/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
+++ b/symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/model/Message.java
@@ -206,8 +206,14 @@ public class Message {
       }
 
       // check if content is encapsulated in <messageML/> node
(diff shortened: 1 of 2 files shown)


thibauult (approved):
LGTM 👍🏻

PR #860, opened 2026-02-17
UPL-244 Add until parameter to listMessages

tzhao-symphony:


symphony-bdk-core/src/test/java/com/symphony/bdk/core/service/message/MessageServiceTest.java
```diff
@@ -56,6 +56,7 @@
 import org.junit.jupiter.api.io.TempDir;
 import org.mockito.ArgumentCaptor;
 
+import java.awt.*;
```
FabienVSymphony: You should not have glob imports in this project
Why is AWT used? this is a graphical library and some of it's dependencies can cause issues on headless environments. I'd avoid to depend on awt if possible

symphony-bdk-core/build.gradle
```diff
@@ -76,7 +76,7 @@ dependencies {
 }
 
 // OpenAPI code generation
-def apiBaseUrl = "https://raw.githubusercontent.com/finos/symphony-api-spec/5526c5e81cb2313c4aab18119b526534652ba98e"
+def apiBaseUrl = "https://raw.githubusercontent.com/finos/symphony-api-spec/<TO BE UPDATED>>"
```
FabienVSymphony: If you want to test you can point to your fork for now

symphony-bdk-core/src/main/java/com/symphony/bdk/core/service/message/MessageService.java
```diff
@@ -42,6 +42,7 @@
 
 import lombok.extern.slf4j.Slf4j;
 import org.apiguardian.api.API;
+import org.jetbrains.annotations.NotNull;
```
FabienVSymphony: Please avoid including jetbrains specific annotations to be compatible on all envs

FabienVSymphony (approved):

FabienVSymphony:
Github action is bugged, there is not dependency change on this PR, so I'll merge anyway.

PR #870, opened 2026-05-04
Update jjwt version and bump Spring version

symphony-enrico:
### Description
update very old jjwt version to the latest one


symphony-enrico:
ok to test

thibauult (approved):

PR #888, opened 2026-08-05
Modernize build toolchain: Gradle 9, bytecode-processing tooling bumps

thibauult:
## Summary

Moves the build itself onto current tooling before the BDK 4.x branch (`next`) baselines Java 25. None of this was a problem on JDK 17 — it becomes a hard blocker the moment the target JDK moves, since every bytecode-processing tool in the build carries a class-file-version floor. Doing it here keeps this mechanical, low-risk diff separate from the riskier Spring Boot / JDK migrations that follow, and lets those changes start from an already-modern build.

Deliberately out of scope: Java baseline (stays 17), Spring Boot version (stays 3.5.16), OpenAPI generator version (stays 6.6.0). Each of those gets its own change (see `openspec/changes/`).

- **Gradle 8.14.5 → 9.6.1**, plus the deprecation cleanup it forces: the removed `sourceCompatibility` convention replaced with an explicit Java 17 toolchain in `bdk.java-common-conventions`, `project.buildDir` → `layout.buildDirectory`, `tasks.create` → `tasks.register`, cross-project `sourceSets` access fixed in both Spring Boot starter modules, and the two Spring Boot Gradle plugin versions in the example modules bumped to match the BOM.
- **Bytecode-processing tooling raised** to versions supporting current class file versions: `byte-buddy` 1.12.19 → 1.18.11, `mapstruct`/`mapstruct-processor` 1.4.2.Final → 1.6.3, `archunit-junit5` 1.2.1 → 1.5.0, plus an explicit `jacoco.toolVersion`. The MapStruct bump produces no behavior change in the generated `UserDetailMapperImpl` (diffed before/after).
- **`mockserver-netty` 5.15.0 → 7.5.0.** Checked first whether a replacement (WireMock/MockWebServer) was needed since Spring Boot 4 also moves Netty — it wasn't; 7.5.0's own POM already targets Netty 4.2. Full test suite passes unmodified across the jump.
- **Lombok pinned explicitly in `symphony-bdk-bom`** (1.18.46, same version currently resolved transitively) instead of inheriting from `spring-boot-dependencies`, so a future Spring Boot bump can't move it out from under the 13 modules that use it.
- **Dead dependencies removed**: `javax.xml.bind:jaxb-api:2.3.1` (zero source usage) and `org.projectreactor:reactor-spring:1.0.1.RELEASE` (2017 artifact, no module depends on it). `javax.annotation:jsr250-api` was investigated for the same treatment but is load-bearing — openapi-generator's jersey2 generator emits `@javax.annotation.Generated` across 335+ generated classes and `jakarta.annotation-api` doesn't provide that package — so it was left as-is.
- **JDK 25 forward-compatibility spike** (throwaway branch, not merged): with all the above bumps in place, flipping the toolchain to Java 25 and running the full build/test/coverage suite produced zero failures. Recorded in `openspec/changes/modernize-build-toolchain/tasks.md` so the follow-up `adopt-java-25-baseline` change doesn't go looking for failures that don't exist.

Removing a BOM constraint (`reactor-spring`) and adding one (`lombok`) are consumer-visible dependency-resolution changes, so this ships as **3.6.0**, not a patch.

## Test plan
- [x] `./gradlew clean build --warning-mode all` — warning-free
- [x] `./gradlew build jacocoTestReport jacocoTestCoverageVerification` — all pass, no threshold adjustments
- [x] `./gradlew publishToMavenLocal` — verified
- [x] Full test suite in `symphony-bdk-core` and `symphony-bdk-http-webclient` re-run unmodified across the mockserver-netty jump
- [x] Generated `UserDetailMapperImpl` diffed before/after the MapStruct bump (cosmetic only)
- [x] JDK 25 toolchain spike run on a throwaway branch with all tooling bumps in place


symphony-enrico (approved):
LGTM

PR #890, opened 2026-08-11
fix: restore missing datafeed includeInvisible flag logic

symphony-enrico:
This commit fixes a regression introduced in 22d25f0 where the DatafeedLoopV2 agent inadvertently stopped propagating the `includeInvisible` configuration flag to the datafeed creation request.

The flag's absence caused events from invisible rooms to be silently filtered out for the entire lifetime of the created feed. The same commit also broke the tests meant to safeguard this logic by changing their mock expectations to expect the flag's absence.

- Restores the propagation of `includeInvisible` from BdkDatafeedConfig to the V5DatafeedCreateBody inside DatafeedLoopV2.
- Reverts all 17 mock expectations in DatafeedLoopV2Test back to enforcing the presence of `includeInvisible(true)` (which is the configured default in the test config.yaml).
- Adds a missing `Consumer` import to DatafeedLoopV2Test to fix the resulting build failure.


thibauult (approved):
LGTM 👍🏻

PR #893, opened 2026-08-11
Propose OpenSpec change: JDK HttpClient-based ApiClient transport

thibauult:
## Summary
- Proposes a new `symphony-bdk-http-jdk` module implementing the BDK `ApiClient`/`ApiClientBuilder` contract on top of `java.net.http.HttpClient`, giving consumers a zero-third-party-dependency alternative to the existing jersey2/webclient implementations.
- Full behavioral parity with `symphony-bdk-http-jersey2` (param encoding, JSON serialization, tracing headers, file downloads, TLS/proxy config, retry-compatible exception translation), with two documented, deliberate gaps (read-timeout mapping, filter chain narrowing) called out explicitly.
- Purely additive: no changes to `symphony-bdk-http-api` contracts or existing modules; Spring Boot starter wiring is explicitly out of scope for this change.

This is an OpenSpec proposal only (`openspec/changes/jdk-httpclient-transport/`) — no implementation code yet.

## Test plan
- [x] Review proposal/design/spec/tasks for completeness and alignment with existing module patterns
- [x] On approval, implementation will follow via `openspec-apply-change`


symphony-enrico (approved):
LGTM !

